PR #54, opened 2025-07-20
Refactor LifecycleFlag to help editors

tomschr:
Dynamically created class didn't work well with VSCode and attribute access.

* Use a real class
* Remove `BaseLifecycleFlag`. It's not needed anymore
* Use `auto()` to get integer value
* Use only "unknown" in lowercase to avoid problems with iterating over the class


tomschr:
> Thanks for the PR @tomschr. This looks fine to me.

Thanks, I'll merge it.

Issue #36, opened 2025-07-10
Add MacOS runner to GitHub Action


tomschr:
### Situation

Currently, our GitHub Action only uses Ubuntu as a runner.

### Use Case

Having a MacOS runner would help those people who work on a Mac.

### Possible Implementation

The easiest way (which don't work!) would be to amend the GitHub Action `ci.yml`:

<details>
  <summary>Diff of the GHA workflow file (click me!)</summary>

```diff
diff --git a/.github/workflows/ci.yml b/.github/workflows/ci.yml
index f31be28..0f00fac 100644
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,16 +16,24 @@ permissions:
 
 jobs:
   test:
-    name: python
-    runs-on: ubuntu-latest
+    name: python (${{ matrix.os }})
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest]
 
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
 
-      - name: Install tools
+      - name: Install tools (Ubuntu)
+        if: matrix.os == 'ubuntu-latest'
         run: sudo apt-get update && sudo apt-get install -y jing libxml2-utils xsltproc
 
+      - name: Install tools (macOS)
+        if: matrix.os == 'macos-latest'
+        run: brew install jing libxml2-utils xsltproc
+
```

</details>

Unfortunately, it fails due to `jing` not available over `brew`. According to Gemini, it was recently removed from "Homebrew's core repository because the upstream project is archived."

Possible options

* **Option 1: The Quick Fix (Install from URL)**
  For a fast, one-off installation on your machine, you can tell Homebrew to install jing using the URL of its last-known formula file from the homebrew-core repository's history. See [Homebrew/homebrew-core: `Formula/j/jing-trang.rb`](https://github.com/Homebrew/homebrew-core/blob/main/Formula/j/jing-trang.rb)

* **Option 2: Create a Custom Tap**
  The best practice for managing project-specific or deprecated formulas is to create your own "tap". A tap is just a Git repository that contains Homebrew formulas. This makes installation clean and simple for anyone working on the project. You need to create a separate GitHub repo. It's not recommended to put it inside this one (although technically possible, but due to separation of concerns we should separate it)

* **Option 3: Avoid jing**
  This is discussed in the attached sub-issue. Rely on `lxml` validation and don't use `jing` at all.

PR #76, opened 2025-08-01
Add MacOS Runner to GitHub Actions

sushant-suse:
1. **Update in the `.github/workflows/ci.yml` file**:
  Updates the CI pipeline configuration by adding a new job to run tests on macOS alongside the existing Ubuntu job. This ensures cross-platform compatibility and increases the coverage of the CI process by testing on both macOS and Ubuntu. The updated `ci.yml` now supports testing on multiple platforms, making the CI process more robust and reliable. Additionally, I have added detailed comments to each step in the configuration for better clarity and maintainability.
2. **Update in the `tests/cli/cmd_repo/test_cmd_list.py` file**:
  Fixed an issue with the formatting of command output in the test by replacing newline characters (`\n`) with spaces. This ensures that the command output is correctly formatted for comparison in the test assertions. The change improves the accuracy of the test by aligning the expected output format with the actual output.
3. **Update in the `tests/utils/test_contextmgr.py` file**:
  Updated the test assertions to modify the `abs` tolerance in `pytest.approx` to `0.2`. This adjustment was made to better accommodate slight timing variations between the measured and expected elapsed time values. The new tolerance ensures that the test is less sensitive to small timing discrepancies, improving test reliability without compromising the validation logic.

Related Issue: #36

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,5 +1,4 @@
 name: CI/Test
-
 on:
   workflow_dispatch:
   pull_request:
@@ -9,27 +8,23 @@ on:
     paths:
       - 'src/**'
       - 'tests/**'
-
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
-
 permissions:
   contents: read
   packages: read
 
 jobs:
-  test:
-    name: python
+  test-ubuntu:
+    name: ubuntu-python
     runs-on: ubuntu-latest
     container:
       image: registry.opensuse.org/documentation/containers/15.6/opensuse-daps-toolchain:latest
       # image: ghcr.io/opensuse/docbuild/daps:latest
-
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
-
       - name: Get dependencies
         run: |
           rpm -q daps \
@@ -50,7 +45,6 @@ jobs:
             jq \
             rsvg-convert \
             openssh-clients || true
-
       - name: Install the latest version of uv
         id: setup-uv
         uses: astral-sh/setup-uv@v6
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
tomschr, line 1: [Add a news fragment file](https://opensuse.github.io/docbuild/developer/add-newsfragments.html) to this PR. I guess, the `infra` type would be sufficient.

It doesn't have to be much, just mentioning that MacOS is added to the test CI.

.github/workflows/ci.yml
```diff
@@ -61,10 +55,65 @@ jobs:
...
+          # Define the version and URL for jing.jar
+          JING_VERSION="20091111"
+          JING_URL="https://repo1.maven.org/maven2/com/thaiopensource/jing/20091111/jing-20091111.jar"
+
+          # Download jing.jar file for XML validation
+          curl -sL -f $JING_URL -o jing.jar
+          
+          # Ensures the jar file is successfully downloaded, else exit with error
+          if [ ! -f jing.jar ]; then
+              echo "Error: Failed to download jing.jar"
```
tomschr, line 100: I'd recommend to use the error message from the [GitHub docs](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands#example-creating-an-annotation-for-an-error):

```suggestion
              echo "::error file=$JING_URL:: Failed to download jing.jar"
```

This will print the error in red.

.github/workflows/ci.yml
```diff
@@ -61,10 +55,65 @@ jobs:
...
+        with:
+          python-version: '3.12'
+      
+      # Install uv and project dependencies inside a virtual environment
+      - name: Install uv and dependencies
+        run: |
+          python3 -m venv .venv
+          source .venv/bin/activate
+          pip install uv
+          uv pip install --editable . --group devel
```
tomschr, line 84: Does the `astral-sh/setup-uv@v6` step doesn't work? 

```yaml
- name: Install the latest version of uv
  id: setup-uv
  uses: astral-sh/setup-uv@v6
  with:
      version: "latest"
      enable-cache: true
      cache-suffix: "docbuild"
```

That would be consistent with the Linux variant. 

However, it's a different approach:

* The Linux test install `uv` and the tool creates the virtual environment. The `uv` tool is _outside_ of the virtual environment. Furthermore, `uv` itself creates the virtual environment.
* The MacOS test creates a virtual environment through `pip` and install `uv` also through `pip`. The `uv` tool is _inside_ the virtual environment.

I'm not sure how similar they are. It may lead to subtle differences. If feasible, I would try to make the steps mostly the same to guarantee the same environment and the same level of reproducibility.

I would also suggest to make the steps a bit smaller. This helps to have smaller, more separate steps which are easier to debug ("separation of concerns").

For example, the step about installing uv and installing the dependencies. They should be separate in my humble opinion:

```suggestion
      # Install uv and project dependencies inside a virtual environment
      - name: Install uv
        id: setup-uv-macos
        run: |
          python3 -m venv .venv && source .venv/bin/activate
          pip install uv
          uv_version=$(uv --version)
          echo "uv_version=$uv_version" >> "$GITHUB_OUTPUT"
 
      - name: Print the installed version
        run: echo "Installed uv version is ${{ steps.setup-uv-macos.outputs.uv_version }}"

      - name: Install dependencies
        run: uv sync --frozen --group devel
```

tomschr (commented):
Hi @sushant-suse,
thanks a lot for your efforts! This looks good!

My only concern is how reproducible is this? I've tried to describe that in one of my comments below.

If feasible, I would like to prefer the steps for Linux and MacOS mostly the same. It's makes it easier to reproduce the steps and to maintain the workflow file. But probably we don't have much time for this.

Maybe it's better to use it with the small suggestions, hope for the best, and investigate more into it later. We have higher and more important priorities than this.

What do you think? 🙂

.github/workflows/ci.yml
sushant-suse: Added, thanks for the reminder.

.github/workflows/ci.yml
```diff
@@ -61,10 +55,65 @@ jobs:
...
+      
+      # Install macOS-specific external tools (such as XML parsers and Java)
+      - name: Install external tools (macOS)
+        run: |
+          brew install libxml2 libxslt openjdk
+
+          # Define the version and URL for jing.jar
+          JING_VERSION="20091111"
+          JING_URL="https://repo1.maven.org/maven2/com/thaiopensource/jing/20091111/jing-20091111.jar"
```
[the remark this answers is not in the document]
sushant-suse: Thanks for the valuable suggestions. I have incorporated the changes.

.github/workflows/ci.yml
sushant-suse: Modified, thanks.

Issue #27, opened 2025-07-07
Allow ID/IDREF checks for validation


tomschr:
### Situation

Currently, we rarely have the ID datatype in the RNG schema (see `src/docbuild/config/xml/data/product-config-schema.rnc`). However, that might change in the future.

### Use Case

To be prepared for having validation with and without ID/IDREF checks, we need to enable the validation and use `-i` when we validate with `jing`. This will help to spot problems in our XML config files.

### Possible Implementation

* Implement the diff (see below)
* Amend the test suite accordingly.
* Research for lxml and activating ID/IDREF check

<details>
   <summary>Diff for `process_validation.py`</summary>

```diff
diff --git i/src/docbuild/cli/process_validation.py w/src/docbuild/cli/process_validation.py
index 59d69e9..8130be0 100644
--- i/src/docbuild/cli/process_validation.py
+++ w/src/docbuild/cli/process_validation.py
@@ -105,6 +105,7 @@ async def validate_rng(
     rng_schema_path: Path = PRODUCT_CONFIG_SCHEMA,
     *,
     xinclude: bool = True,
+    idcheck: bool = True,
 ) -> tuple[bool, str]:
     """Validate an XML file against a RELAX NG schema using jing.
 
@@ -116,9 +117,12 @@ async def validate_rng(
     :param rng_schema_path: The path to the RELAX NG schema file. It supports
         both RNC and RNG formats.
     :param xinclude: If True, resolve XIncludes with `xmllint` before validation.
+    :param idcheck: If True, perform ID uniqueness checks.
     :return: A tuple containing a boolean success status and any output message.
     """
     jing_cmd = ['jing']
+    if idcheck:
+        jing_cmd.append('-i')
     if rng_schema_path.suffix == '.rnc':
         jing_cmd.append('-c')
     jing_cmd.append(str(rng_schema_path))
```
</details>

PR #80, opened 2025-08-31
bugfix: Fixes incorrect `jing` command arguments in `validate_rng`

sushant-suse:
Linked issue: #27 

This change resolves the bug by ensuring the `'-i'` flag is only added to the command when the `idcheck` parameter is explicitly `True`, aligning the function's behaviour with its intended purpose and the test's expectations.

#### 1. In `src/docbuild/cli/cmd_validate/process.py` file:

- **Change**: Modified the logic for appending the `'-i'` flag to the `jing_cmd` list.
- **Reason**: The previous logic added the `'-i'` flag unconditionally if `idcheck` was `True` (which is its default value), causing an `AssertionError` in the test case where `xinclude=False` and `idcheck` was not explicitly passed. This fix ensures the `'-i'` flag is only added to the command when `idcheck` is explicitly set to `True`, which aligns the function's behaviour with the expected test output.

#### 2. In `tests/cli/cmd_validate/validate/test_process_validation.py` file:

- **Change**: Modified the `test_validate_rng_jing_failure` test case.
- **Reason**: The test was failing because its `assert_called_once_with` statement expected the `jing` command to be called without the `'-i'` flag. The test was updated to explicitly pass `idcheck=False` to the `validate_rng` function. This change allows the test to pass correctly and ensures the test's input parameters match its own expectations for the command call.

#### 3. In `tests/cli/cmd_validate/test_process.py` file:

- **Change**: Added three new test cases to ensure proper functionality of the `idcheck` parameter within the `validate_rng` function.
- **Reason**: The original bug fix highlighted a lack of test coverage for the `idcheck` feature. These new tests verify that:
    - The function correctly passes the `-i` flag to `jing` when `idcheck=True` and succeeds on a valid file.
    - The function correctly passes the `-i` flag and fails when a duplicate ID is present in the XML.
    - The function correctly does not pass the `-i` flag when `idcheck=False`, and validation succeeds even if a duplicate ID would otherwise cause a failure.

### DIFF
--- a/changelog.d/80.bugfix.rst
+++ b/changelog.d/80.bugfix.rst
@@ -1,0 +1,1 @@
+Fixes incorrect jing command arguments in validate_rng function.
--- a/src/docbuild/cli/cmd_validate/process.py
+++ b/src/docbuild/cli/cmd_validate/process.py
@@ -106,6 +106,7 @@ async def validate_rng(
     rng_schema_path: Path = PRODUCT_CONFIG_SCHEMA,
     *,
     xinclude: bool = True,
+    idcheck: bool = True
 ) -> tuple[bool, str]:
     """Validate an XML file against a RELAX NG schema using jing.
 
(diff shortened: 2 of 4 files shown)


sushant-suse:
> @patch.object(process_module, 'run_command', new_callable=AsyncMock)
> @pytest.mark.asyncio

Thanks for the feedbacks, I have added the changes

sushant-suse:
> Thank you @sushant-suse! Looks fine! I have only three minor things, otherwise it's great. 👍
> 
> Make sure to add a `Fix #27: ...` message in your Git commit log. That closes the issue automatically.

Thanks for the feedbacks, I have added the changes

tomschr (approved):

Issue #63, opened 2025-07-22
Make the GitHub Action release workflow more robust


tomschr:
### Situation

Currently, you create a release branch and name a version to it (e.g. `release/1.0.0`). However, we don't check the version of the release with the version of the tool.

### Use Case

We avoid user errors when we check, if the version of the release branch is equal to the version of the tool (`src/docbuild/__about__.py`).

### Possible Implementation

For the release workflow (`.github/workflows/release.yml`) I would suggest:

1. Extract the version from the release branch.
2. Extract the version from the tool. There are actually two ways to do it:
   1. Directly by looking at the `__version__` variable in the file `src/docbuild/__about__.py`.
   2. Indirectly by calling the command `uv run docbuild --version`. That's probably a bit more portable and independent where the version is stored. But it also needs a building step. 🤔 
4. Compare the too versions. If they differ, reject the merge and offer a meaningful error message.

sushant-suse:
Solved in PR #86

PR #86, opened 2025-10-07
Feat #63: Add version consistency check to release workflow

sushant-suse:
Related Issue #63 

### Changes made

Added a new automated check to the `release.yml` GitHub Actions workflow.

- A new step named "Check Version Consistency" has been added to the `do-release` job. This step extracts the version number from the release branch name and compares it to the `__version__` string defined in `src/docbuild/__about__.py`.

### Why changed above?

The purpose of this change is to prevent human errors that can lead to version mismatches and broken releases.

- The check acts as an automated quality gate, ensuring that the version number in the Git branch name (for example, `release/1.2.3`) always matches the version number in the codebase.
- If a developer pushes a release branch with a version that does not match the code, the CI will now fail immediately. This prevents an invalid Git tag from being created and stops the release process before it can cause issues for users.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -53,8 +53,8 @@ jobs:
           # Safely extract __version__ from the file (for example, '1.0.0' from __version__ = '1.0.0')
           CODE_VERSION=$(grep '__version__' src/docbuild/__about__.py | awk -F "'" '{print $2}')
           
-          echo "Branch version: $BRANCH_VERSION"
-          echo "Code version:   $CODE_VERSION"
+          echo ":notice title=Branch version::$BRANCH_VERSION"
+          echo ":notice title=Code version::$CODE_VERSION"
 
           if [ "$BRANCH_VERSION" != "$CODE_VERSION" ]; then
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -45,6 +45,24 @@ jobs:
...
+      - name: Check Version Consistency
+        id: version_check
+        run: |
+          BRANCH_VERSION="${{ env.VERSION }}"
+          
+          # Safely extract __version__ from the file (for example, '1.0.0' from __version__ = '1.0.0')
+          CODE_VERSION=$(grep '__version__' src/docbuild/__about__.py | awk -F "'" '{print $2}')
+          
+          echo "Branch version: $BRANCH_VERSION"
+          echo "Code version:   $CODE_VERSION"
```
tomschr, line 57: I would use a "notice" command from the [Workflow commands](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands#setting-a-notice-message) page:


Syntax:
```
::notice file={name},line={line},endLine={endLine},title={title}::{message}
```

```suggestion
          echo ":notice title=Branch version::$BRANCH_VERSION"
          echo ":notice title=Code version::$CODE_VERSION"
```

tomschr (changes requested):
Thanks @sushant-suse! Just a minor thing to improve logging in the GitHub workflow.

PR #89, opened 2025-10-22
Fix #82: Implement environment-specific concurrency lock (PID file)

### LINKED ISSUE #82: How do we make the script idempotent?
### Definition

"**Idempotence** is the property of certain operations in mathematics and computer science whereby they can be applied multiple times without changing the result beyond the initial application." —[Source](https://en.wikipedia.org/wiki/Idempotence)

### Situation

Let's assume, we started the `docbuild` script in one terminal and an additional instance in another terminal.

How should the second `docbuild` behave? Several options are possible:

* Immediately stop. Allow only one instance to run.
* Try to make the internal state idempotent. That could be hard, but necessary if we want to have this as a backend in the future.
* If it's about building, add the additional task to the queue of the first `docbuild` process and let the second instance return immediately.
* Figure out, which sub commands can be run without messing up the internal state. Usually these are commands that only queries things.

### Use Case

Determining whether there is another `docbuild` process is crucial. A second instance shouldn't mess up the internal state

### Possible Implementation

The harder parts are still to be discussed. But checking about another instance could be done like this: 
 
* Check if there the file `/var/run/docbuild.pid` exists.
* If that's not the case, store the current PID in this file.
* If we already have a running instance, we need to look at the instances and compare what env configs are currently being used:
  * Two instances using the same env config. If they do, then don't let run the instance that was started later.
  * Using different env configs. The later started instance is allowed to run.

sushant-suse:
Related Issue #82

### Changed made

* **New Utility Class (`src/docbuild/utils/pidlock.py`)**: Implemented the `PidFileLock` class, which manages environment-specific lock files.
    * The lock file name is generated using a secure hash (SHA256) of the absolute environment configuration file path, ensuring uniqueness per configuration.
    * The utility includes logic to check the status of the process in the PID file, automatically cleaning up stale lock files (files belonging to a dead process).
    * It uses `atexit.register` and `atexit.unregister` to ensure the lock is reliably released both on normal program termination and manual context exit.
* **CLI Integration (`src/docbuild/cli/cmd_cli.py`)**: Integrated the locking mechanism into the main `cli` function.
    * The application now determines the canonical path of the loaded environment config file.
    * It attempts to acquire the `PidFileLock` using this path as the unique resource ID.
    * If the lock is held by another running process, the application logs an error and exits with code 1, preventing execution.
* **Constant Definition (`src/docbuild/constants.py`)**: Added the `BASE_LOCK_DIR` constant to define the standardized, user-specific location for lock files.
* **Test Suite**: Added a comprehensive suite of unit tests for `PidFileLock` (`tests/utils/test_pidlock.py`) that mock OS signals (`os.kill`) to verify all scenarios, including stale lock cleanup and concurrency refusal.

### Why changed above?

1.  **Preventing State Corruption**: When two instances of `docbuild` attempt to modify the same environment (for example, writing to the same cache or modifying the same files), race conditions occur, leading to corrupted output and unpredictable behavior. The lock ensures that only one process can work with a specific environment ID at any given time.
2.  **Granular Concurrency Control**: Unlike a simple application-wide lock, this feature is environment-specific. This means a user is allowed to run:
    * `docbuild --env-config env.devel.toml` (Lock A)
    * `docbuild --env-config env.test.toml` (Lock B)
    Simultaneously, but cannot run two processes against the same file.
3.  **Reliable Cleanup**: The lock management is robust:
    * **Acquisition Failure**: If a running process holds the lock, a clear `RuntimeError` is raised.
    * **Stale Lock Fix**: The system prevents perpetual locks by automatically checking if the PID in the lock file is still active, removing the file if the original process is dead.
4.  **Best Practice Implementation**: Using a PID file is the standard Unix-idiomatic way to manage concurrency for long-running user processes. 

### DIFF
--- a/src/docbuild/utils/pidlock.py
+++ b/src/docbuild/utils/pidlock.py
@@ -5,6 +5,7 @@ import hashlib
 import logging
 import os
 import errno
+import os
 from pathlib import Path
 from typing import Self, Any
 
@@ -18,7 +19,7 @@ class PidFileLock:
 
     The lock file is named based on a hash of the environment config file path.
     """
-
+    
     def __init__(self, resource_path: Path, lock_dir: Path = BASE_LOCK_DIR) -> None:
         """Initialize the lock manager.
 
@@ -26,13 +27,26 @@ class PidFileLock:
         :param lock_dir: The base directory for lock files.
         """
         self.resource_path = resource_path.resolve()
-        self.lock_dir = lock_dir
-        # The lock_path is already calculated here, fixing the AttributeError in test_acquire_and_release_success
-        self.lock_path: Path = self._generate_lock_name() 
+        
+        # Converted public attributes to private attributes
+        self._lock_dir: Path = lock_dir
+        self._lock_path: Path = self._generate_lock_name()
+        
         self._lock_acquired: bool = False
 
     @property
-    def lock(self):
+    def lock_dir(self) -> Path:
+        """The directory where PID lock files are stored."""
+        return self._lock_dir
+
+    @property
+    def lock_path(self) -> Path:
+        """The full path to the lock file."""
+        return self._lock_path
+    
+    @property
+    def lock(self) -> bool:
+        """Return the current lock acquisition state."""
         return self._lock_acquired
     
     def __enter__(self) -> Self:
@@ -50,15 +64,14 @@ class PidFileLock:
         """Generate a unique lock file name based on the resource path."""
         # SHA256 hash of the absolute path is used to ensure a unique, safe filename.
         path_hash = hashlib.sha256(str(self.resource_path).encode('utf-8')).hexdigest()
-        return self.lock_dir / f'docbuild-{path_hash}.pid'
+        return self._lock_dir / f'docbuild-{path_hash}.pid'
 
     def acquire(self) -> None:
         """Acquire the lock atomically, or diagnose an existing lock."""
         self.lock_dir.mkdir(parents=True, exist_ok=True)
-        self.lock_path = self._generate_lock_name()
         current_pid = os.getpid()
 
-        # 1. Attempt to acquire the lock file atomically (O_CREAT | O_EXCL)
+        # 1. Attempt to acquire the lock file atomically (O_CREAT | os.O_EXCL)
         try:
             # os.O_WRONLY | os.O_CREAT | os.O_EXCL ensures file is created ONLY if it doesn't exist.
             # 0o644 sets the file permissions (read/write for owner, read for others).
@@ -125,18 +138,16 @@ class PidFileLock:
 
     def release(self) -> None:
         """Release the lock file."""
-        if self._lock_acquired and self.lock_path and self.lock_path.exists():
+        if self._lock_acquired and self.lock_path.exists():
             try:
                 self.lock_path.unlink()
                 self._lock_acquired = False
                 log.debug("Released lock at %s.", self.lock_path)
                 
                 # Unregister the cleanup function
-                # This must be done to prevent the lock from being released twice
-                # (once manually, once at program exit)
                 atexit.unregister(self.release)
             except OSError as e:
-                log.error("Failed to remove lock file at %s: %s", (self.lock_path, e))
+                log.error(f"Failed to remove lock file at {self.lock_path}: {e}")
 
     @staticmethod
     def _is_pid_running(pid: int) -> bool:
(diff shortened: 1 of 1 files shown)


src/docbuild/utils/pidlock.py
```diff
@@ -0,0 +1,131 @@
...
+            except FileNotFoundError:
+                # File was removed between exists() and open(). Continue.
+                pass
+            except OSError as e: 
+                # Catching OSError for general I/O problems during read/unlink
+                log.error(f"Non-critical error while checking lock file: {e}")
+                pass 
+            except ValueError:
+                # Catching ValueError if the PID file content is not an integer
+                log.warning(f"Lock file at {self.lock_path} contains invalid PID. Removing.")
```
tomschr: Same here:


```suggestion
                log.warning("Lock file at %s contains invalid PID. Removing.", self.lock_path)
```

src/docbuild/utils/pidlock.py
```diff
@@ -0,0 +1,131 @@
...
+        """Acquire the lock or raise an error if another instance is running."""
+        self.lock_dir.mkdir(parents=True, exist_ok=True)
+        current_pid = os.getpid()
+
+        # 1. Check for stale lock
+        if self.lock_path.exists():
+            try:
+                # Read PID from the file
+                with self.lock_path.open('r') as f:
+                    pid_str = f.read().strip()
```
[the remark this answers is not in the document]
sushant-suse: > Just a wild idea: could we add the PID inside the filename? 🤔 This would avoid reading the file _content_ and we could instantly check that.
> 
> Not sure if it's good, but I think we can discuss this.

Given that the current logic requires the lock file to be unique per environment, and the PID is only known after the lock is acquired, placing the PID in the filename isn't suitable for generating the lock path before checking for an existing lock.

However, we can simplify the process by removing the need to read the PID file entirely, using a more standard locking technique: Atomic File Creation.

Instead of reading the PID, we can rely on the OS to handle the lock acquisition atomically. This is done using a low-level function like `os.open()` with the `os.O_CREAT` and `os.O_EXCL` flags, which ensures that the lock file is created only if it does not already exist. This check is instantaneous and avoids race conditions.

By using this method, we no longer check for a running PID by reading the file content first. Instead, we use the lock's existence as the flag.

Maybe we can discuss it later in a new issue?

tests/utils/test_pidlock.py
```diff
@@ -0,0 +1,164 @@
...
+    resource_path, lock_dir = lock_setup
+    
+    current_pid = os.getpid()
+
+    lock = PidFileLock(resource_path, lock_dir)
+    assert not lock.lock_path.exists()
+    
+    # Acquire lock via context manager
+    with lock:
+        assert lock._lock_acquired is True
```
tomschr: After adding the property, we should use it. 😉 


```suggestion
        assert lock.lock is True
```

tests/utils/test_pidlock.py
```diff
@@ -0,0 +1,164 @@
...
+    """Test acquiring the lock when the lock directory does not yet exist."""
+    resource_path, lock_dir = lock_setup
+    
+    # Ensure lock_dir is deleted if it exists
+    if lock_dir.exists():
+        shutil.rmtree(lock_dir)
+
+    lock = PidFileLock(resource_path, lock_dir)
+    with lock:
+        assert lock._lock_acquired is True
```
tomschr: Same here:


```suggestion
        assert lock.lock is True
```

tomschr (changes requested):
@sushant-suse Thanks. Just found some `._lock_acquired` attributes being used in tests. We should use `.lock` consistently instead of using the "private" attribute name.

src/docbuild/utils/pidlock.py
```diff
@@ -0,0 +1,156 @@
...
+    def __init__(self, resource_path: Path, lock_dir: Path = BASE_LOCK_DIR) -> None:
+        """Initialize the lock manager.
+
+        :param resource_path: The unique path identifying the resource to lock (e.g., env config file).
+        :param lock_dir: The base directory for lock files.
+        """
+        self.resource_path = resource_path.resolve()
+        self.lock_dir = lock_dir
+        # The lock_path is already calculated here, fixing the AttributeError in test_acquire_and_release_success
+        self.lock_path: Path = self._generate_lock_name()
```
tomschr: Sorry, I forgot these two attributes. I would convert them also into properties. Also use the property name in the tests.

tomschr (changes requested):

PR #90, opened 2025-10-22
Refactor: updated logo and added favicon

sushant-suse:
Updated Logo and added the logo as favicon.


docs/source/_static/logo.png
tomschr: For the logo 1024px is good, but there are some issues if you use this logo as a favicon:

* **Size** The logo is 1024 ⨉ 1024. That's too big for a favicon. Usually they are small in size and used mostly in browser tabs. Think of having such big favicon on many browser tabs. It can slow down your browser.
* **Format** Although PNG would work, the traditional format is ICO (file extension `.ico`). It can combine different sizes into one file and the browser selects the appropriate one.

I would recommend to use ImageMagick and convert the logo into a favicon:

```
# For ImageMagick version 7+
magick logo.png -define icon:auto-resize="128,64,48,32,16" favicon.ico
```

docs/source/_static/logo-old.png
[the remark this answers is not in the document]
sushant-suse: > I think you can remove this old logo. If we need the old version, we can recover it from Git.

Done

docs/source/_static/logo.png
sushant-suse: > For the logo 1024px is good, but there are some issues if you use this logo as a favicon:
> 
> * **Size** The logo is 1024px ⨉ 1024px. That's too big for a favicon. Usually they are small in size and used mostly in browser tabs. Think of having such big favicon on many browser tabs. It can slow down your browser.
> * **Format** Although PNG would work, the traditional format is ICO (file extension `.ico`). It can combine different sizes into one file and the browser selects the appropriate one.
> 
> I would recommend to use ImageMagick and convert the logo into a favicon:
> 
> ```
> # For ImageMagick version 7+
> magick logo.png -define icon:auto-resize="128,64,48,32,16" favicon.ico
> ```
> 
> If I convert the logo, the `favicon.ico` has a size of 99678 bytes in comparison to 1016983 bytes for the PNG. If you remove the `128` option from `magick` you can reduce the size even more (to 32038 bytes).

Great, thanks for the information. I have never heard about magick. I used it just now and its awesome.
Thanks again.

tomschr (approved):
Thanks! 👍 Please squash-merge.

PR #91, opened 2025-10-23
Fix #88: Implement Pydantic validation for App Configuration

### LINKED ISSUE #88: Use Pydantic for validation of app config
### Situation

Currently, the configuration for the docbuild app isn't validated against errors.

### Use Case

Having a mechanism for validating the content of the app config helps in discovering configuration problems like typos or improper types.

### Possible Implementation

This is still to be defined. At the moment, we don't have much configuration options available for the app itself.

sushant-suse:
Related Issue #88

### Changed made

This PR implements Pydantic schema validation for the application configuration (`AppConfig`) by introducing a new modeling and validation layer.

1.  **Strict Logging Schema Implemented:** The generic `App_LoggingConfig` was replaced with a hierarchy of precise Pydantic models (`FormatterConfig`, `HandlerConfig`, `RootLoggerConfig`, etc.). This enforces validation on sensitive nested parameters (e.g., `datefmt`, `style`, `class` alias) that were previously left unchecked, resolving the issue of "superficial validation."
2.  **Model Structure & Placeholder Resolution:**
    * **New File (`src/docbuild/models/config_model/app.py`):** Defines the complete `AppConfig` schema.
    * Implemented a `@model_validator(mode='before')` in `AppConfig` to handle placeholder resolution (e.g., `{APP_NAME}`) on the raw input data before Pydantic performs validation.
3.  **CLI Integration & Error Handling:**
    * **Updated CLI (`src/docbuild/cli/cmd_cli.py`):** The application now validates the raw config using `AppConfig.from_dict()`. Invalid configs are caught, and the application exits gracefully with a detailed error message.
    * **Updated Loader (`src/docbuild/config/load.py`):** Simplified to handle only raw TOML loading, delegating all processing and validation to the `AppConfig` model.

### Why changed above?

* **Eliminated Runtime Crashes (Actionable Errors):** The application now prevents cryptic runtime crashes. Pydantic validates the entire structure at startup, giving the user clear, actionable error messages for typos or incorrect data types instead of failing deep in the application logic.
* **Achieved Strict Validation for Logging:** The previous approach was weak. The new implementation ensures that the highly sensitive logging configuration adheres to a strict schema, guaranteeing that the `setup_logging` function receives compatible parameters.
* **Correctly Handled Placeholders:** Moving `replace_placeholders` into the `@model_validator` ensures that environment variables and internal references are resolved before Pydantic attempts to coerce values into types like `Path` or `HttpUrl`, which is the correct architectural pattern.
* **Future-Proofing:** This complete validation structure is now the foundation for implementing the larger and more complex Environment Configuration (`EnvConfig`) validation in future PRs.

### DIFF
--- a/src/docbuild/utils/pidlock.py
+++ b/src/docbuild/utils/pidlock.py
@@ -4,6 +4,8 @@ import atexit
 import hashlib
 import logging
 import os
+import errno
+import os
 from pathlib import Path
 from typing import Self, Any
 
@@ -17,7 +19,7 @@ class PidFileLock:
 
     The lock file is named based on a hash of the environment config file path.
     """
-
+    
     def __init__(self, resource_path: Path, lock_dir: Path = BASE_LOCK_DIR) -> None:
         """Initialize the lock manager.
 
@@ -25,8 +27,11 @@ class PidFileLock:
         :param lock_dir: The base directory for lock files.
         """
         self.resource_path = resource_path.resolve()
+        
+        # Converted public attributes to private attributes
         self._lock_dir: Path = lock_dir
         self._lock_path: Path = self._generate_lock_name()
+        
         self._lock_acquired: bool = False
 
     @property
@@ -38,102 +43,97 @@ class PidFileLock:
     def lock_path(self) -> Path:
         """The full path to the lock file."""
         return self._lock_path
-
+    
     @property
     def lock(self) -> bool:
         """Return the current lock acquisition state."""
         return self._lock_acquired
-
+    
     def __enter__(self) -> Self:
         """Acquire the lock."""
         self.acquire()
         return self
 
     def __exit__(
-        self,
-        exc_type: type[BaseException] | None,
-        exc_value: BaseException | None,
-        traceback: Any,
+        self, exc_type: type[BaseException] | None, exc_value: BaseException | None, traceback: Any
     ) -> None:
         """Release the lock."""
         self.release()
 
     def _generate_lock_name(self) -> Path:
         """Generate a unique lock file name based on the resource path."""
-        path_hash = hashlib.sha256(str(self.resource_path).encode("utf-8")).hexdigest()
-        return self._lock_dir / f"docbuild-{path_hash}.pid"
+        # SHA256 hash of the absolute path is used to ensure a unique, safe filename.
+        path_hash = hashlib.sha256(str(self.resource_path).encode('utf-8')).hexdigest()
+        return self._lock_dir / f'docbuild-{path_hash}.pid'
 
     def acquire(self) -> None:
         """Acquire the lock atomically, or diagnose an existing lock."""
         self.lock_dir.mkdir(parents=True, exist_ok=True)
         current_pid = os.getpid()
 
+        # 1. Attempt to acquire the lock file atomically (O_CREAT | os.O_EXCL)
         try:
+            # os.O_WRONLY | os.O_CREAT | os.O_EXCL ensures file is created ONLY if it doesn't exist.
+            # 0o644 sets the file permissions (read/write for owner, read for others).
             lock_fd = os.open(
                 self.lock_path, os.O_WRONLY | os.O_CREAT | os.O_EXCL, 0o644
             )
-            with os.fdopen(lock_fd, "w") as f:
-                f.write(f"{current_pid}\n")
+            # Write PID to the atomically created file
+            with os.fdopen(lock_fd, 'w') as f:
+                f.write(str(current_pid) + '\n')
 
             self._lock_acquired = True
-            try:
-                if logging.getLogger().hasHandlers() and not logging._shutdown:
-                    log.debug("Acquired lock for %s (PID %s).", self.resource_path, current_pid)
-            except Exception:
-                pass
-
+            log.debug(f"Acquired lock for {self.resource_path} at PID {current_pid}.")
+            # Registering release ensures the lock is cleaned up when the program exits.
             atexit.register(self.release)
             return
 
         except FileExistsError:
+            # 2. Lock file exists (atomic open failed). Must now check if it's stale or running.
             pass
         except OSError as e:
+            # Catch file system errors during the atomic open operation
             raise RuntimeError(f"Failed to create lock file at {self.lock_path}: {e}")
 
+        # 3. Lock exists. Check if the process is running.
         if self.lock_path.exists():
             try:
-                with self.lock_path.open("r") as f:
+                # Read PID from the file content (slower operation, only done on contention)
+                with self.lock_path.open('r') as f:
                     pid_str = f.read().strip()
+                
                 pid = int(pid_str)
-
+                
+                # Check if the process is actually running
                 if self._is_pid_running(pid):
+                    # Running instance found. Raise the error.
                     raise RuntimeError(
                         f"docbuild instance already running (PID: {pid}) "
                         f"for configuration: {self.resource_path}"
                     )
                 else:
-                    try:
-                        if logging.getLogger().hasHandlers() and not logging._shutdown:
-                            log.warning(
-                                "Found stale lock file at %s (PID %s). Removing and retrying acquisition.",
-                                self.lock_path,
-                                pid,
-                            )
-                    except Exception:
-                        pass
+                    # Stale lock. Attempt to clean up and acquire again.
+                    log.warning(
+                        f"Found stale lock file at {self.lock_path} (PID {pid}). Removing and retrying acquisition."
+                    )
                     self.lock_path.unlink()
-                    return self.acquire()
+                    
+                    # Recursively call acquire() to try and grab the lock immediately.
+                    return self.acquire() 
 
             except FileNotFoundError:
-                return self.acquire()
+                # Race condition: another process removed the stale lock before us. Retry.
+                return self.acquire() 
             except ValueError:
-                try:
-                    if logging.getLogger().hasHandlers() and not logging._shutdown:
-                        log.warning(
-                            "Lock file at %s contains invalid PID. Removing and retrying.",
-                            self.lock_path,
-                        )
-                except Exception:
-                    pass
+                # Invalid PID format. Treat as stale and clean up.
+                log.warning(f"Lock file at {self.lock_path} contains invalid PID. Removing and retrying.")
                 self.lock_path.unlink()
                 return self.acquire()
             except OSError as e:
-                try:
-                    if logging.getLogger().hasHandlers() and not logging._shutdown:
-                        log.error("Non-critical error while checking lock file: %s", e)
-                except Exception:
-                    pass
-
+                # Non-critical I/O error during read/unlink attempt
+                log.error(f"Non-critical error while checking lock file: {e}")
+            
+        # If all checks and retries fail to acquire the lock, raise a final error.
         raise RuntimeError(f"Failed to acquire lock for {self.resource_path} after multiple checks.")
 
     def release(self) -> None:
@@ -142,35 +142,28 @@ class PidFileLock:
             try:
                 self.lock_path.unlink()
                 self._lock_acquired = False
-
-                # Only log if logging system is active
-                try:
-                    if logging.getLogger().hasHandlers() and not logging._shutdown:
-                        log.debug("Released lock at %s.", self.lock_path)
-                except Exception:
-                    pass
-
-                # Unregister the cleanup hook safely
-                try:
-                    atexit.unregister(self.release)
-                except Exception:
-                    pass
-
+                log.debug("Released lock at %s.", self.lock_path)
+                
+                # Unregister the cleanup function
+                atexit.unregister(self.release)
             except OSError as e:
-                try:
-                    if logging.getLogger().hasHandlers() and not logging._shutdown:
-                        log.error(f"Failed to remove lock file at {self.lock_path}: {e}")
-                except Exception:
-                    pass
+                log.error(f"Failed to remove lock file at {self.lock_path}: {e}")
 
     @staticmethod
     def _is_pid_running(pid: int) -> bool:
-        """Check whether a given PID is currently running."""
-        try:
-            os.kill(pid, 0)
-        except ProcessLookupError:
+        """Check if a process with the given PID is currently running.
+        
+        :param pid: The PID to check
+        :return: True, if the process with the given PID is running,
+                 False otherwise
+        """
+        if pid <= 0:
             return False
-        except PermissionError:
+        try:
+            # Sending signal 0 will raise an OSError exception if the pid is
+            # not running. Do nothing otherwise.
+            os.kill(pid, 0)
             return True
-        else:
-            return True
+        except OSError as err:
+            # Check for ESRCH (No such process)
+            return err.errno != errno.ESRCH


src/docbuild/cli/cmd_cli.py
```diff
@@ -136,12 +141,25 @@ def cli(
...
 
-    # Phase 2: Advanced logging setup with user configuration.
-    logging_config = context.appconfig.get('logging', {})
+    # 2. Validate the raw config dictionary using Pydantic
+    try:
+        # Pydantic validation also handles placeholder replacement via @model_validator
+        context.appconfig = AppConfig.from_dict(raw_appconfig)
+    except (ValueError, ValidationError) as e:
+        log.error("Application configuration failed validation:")
+        log.error(f"Error in config file(s): {context.appconfigfiles}")
```
tomschr: Remember lazy logging? 😉 


```suggestion
        log.error("Error in config file(s): %s", context.appconfigfiles")
```

src/docbuild/config/app_model.py
```diff
@@ -0,0 +1,67 @@
...
+    # These attributes are required by Python's logging dictConfig schema
+    version: int = 1
+    disable_existing_loggers: bool = False
+    
+    # We define core sections as dicts to allow any arbitrary content inside,
+    # as defined by the user's logging schema.
+    formatters: dict[str, Any] = Field(default_factory=dict)
+    handlers: dict[str, Any] = Field(default_factory=dict)
+    loggers: dict[str, Any] = Field(default_factory=dict)
+    root: dict[str, Any] = Field(default_factory=dict)
```
tomschr: I would suggest to be a bit more verbose in the `Field` call. You can look for examples in `src/docbuild/models/language.py`.

The reason why I think it's useful is, that the metadata is used when generating the reference documentation. I'd suggest to amend `Field`:

```
    formatters: dict[str, Any] = Field(
        title="Formatters dictionary",
        description="...",
        # frozen=True, # should we do that?
        examples=[
          '...'
        ]
        default_factory=dict
    )
```

Same for the other fields.

src/docbuild/config/app_model.py
```diff
@@ -0,0 +1,67 @@
...
+    This structure is validated for existence of core sections, but allows
+    for the flexible structure required by logging.config.dictConfig.
+    """
+    # These attributes are required by Python's logging dictConfig schema
+    version: int = 1
+    disable_existing_loggers: bool = False
+    
+    # We define core sections as dicts to allow any arbitrary content inside,
+    # as defined by the user's logging schema.
+    formatters: dict[str, Any] = Field(default_factory=dict)
```
tomschr: I'm thinking about the type annotation for `formatters` and the others. 🤔  On one side I feel it's a bit too lax. Logging is tricky, and if you aren't supported then it can break pretty fast.

On the other side if we want to be precise, it leads to more classes. 

If we do it the lax way, what's the point of having a Pydantic validation when validation is only half-way checked? 🤔 

Just an idea (we don't have to implement it like this). What if we create a `FormatterConfig` class like this:

```python
# This model defines the structure for *one* formatter configuration.
# We use this as the *value* in the main 'formatters' dictionary.
class FormatterConfig(BaseModel):
    """
    Defines the configuration for a single logging formatter.
    See: https://docs.python.org/3/library/logging.config.html#logging-config-formatter-dict
    """
    
    # We must allow extra fields ('allow').
    # This is critical because custom formatters (using 'class' or '()')
    # might accept additional, non-standard configuration keys.
    # It also allows the special '()' factory key.
    model_config = ConfigDict(extra='allow')

    # The format string (e.g., "%(asctime)s - %(name)s")
    format: Optional[str] = None
    
    # The date format string (e.g., "%Y-%m-%d %H:%M:%S")
    datefmt: Optional[str] = None
    
    # The style of the format string: '%', '{', or '$'
    style: Optional[Literal['%', '{', '$']] = None
    
    # 'class' is a reserved Python keyword, so we use 'alias'.
    # This field specifies a custom formatter class.
    class_name: Optional[str] = Field(None, alias='class')
    
    # Specifies if the format string should be validated.
    validate: Optional[bool] = None

    # Note: We cannot explicitly model the '()' key as a Pydantic field
    # because '()' is not a valid Python identifier.
    # However, `ConfigDict(extra='allow')` ensures that if a '()' key
    # exists in the config dict, Pydantic will accept it.
```

For the `App_LoggingConfig` class we then could write:

```python
# This is the recommended type for your 'formatters' field.
# It's a dictionary mapping a formatter's name (str) to its
# specific configuration (FormatterConfig).
type FormattersField = dict[str, FormatterConfig]

# Example of how to use this in your main Pydantic model
class MyLoggingConfig(BaseModel):
"""Configuration for the application's logging system.
    
    This structure is validated for existence of core sections, but allows
    for the flexible structure required by logging.config.dictConfig.
    """
    # These attributes are required by Python's logging dictConfig schema
    version: int = 1
    disable_existing_loggers: bool = False
    
    # We define core sections as dicts to allow any arbitrary content inside,
    # as defined by the user's logging schema.
    # The 'formatters' field using our strict type
    formatters: FormattersField | None = None
```

The same would be needed for `handlers`, `loggers`, and `root`. Pretty much code for validation.

src/docbuild/config/app_model.py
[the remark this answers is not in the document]
sushant-suse: > I'm thinking about if the `config` directory is the right place. 🤔 It's a model (although a model for an application configuration). From a consistency point of view, it would be good to have it in one place.
> 
> Maybe we should move it to `docbuild/models/`? What do you think?

Yes, I agree. I have updated it.

src/docbuild/config/app_model.py
sushant-suse: > I would suggest to be a bit more verbose in the `Field` call. You can look for examples in `src/docbuild/models/language.py`.
> 
> The reason why I think it's useful is, that the metadata is used when generating the reference documentation. I'd suggest to amend `Field`:
> 
> ```
>     formatters: dict[str, Any] = Field(
>         title="Formatters dictionary",
>         description="...",
>         # frozen=True, # should we do that?
>         examples=[
>           '...'
>         ]
>         default_factory=dict
>     )
> ```
> 
> Same for the other fields.

Agree, added it.

src/docbuild/config/app_model.py
sushant-suse: > The same would be needed for `handlers`, `loggers`, and `root`. Pretty much code for validation, hmn.

Thanks @tomschr, I completely agree. Using `dict[str, Any]` for the logging sections only provides half-way validation, which compromises the point of introducing Pydantic in the first place. The risk of hard-to-debug runtime errors is too high if a user makes a simple typo like `dateformat` instead of `datefmt`.

I agree that we should adopt the Precise Validation approach by creating dedicated models for `FormatterConfig`, `HandlerConfig`, and `LoggerConfig`.

sushant-suse: I suggest creating a new issue for this maybe we can tackle it later? WDYT?

src/docbuild/models/config_model/app.py
```diff
@@ -1,79 +1,140 @@
...
+    Defines the configuration for a single logging formatter.
+    Allows extra keys for custom class/factory arguments.
+    """
+    # Allows for custom factory parameters and the special '()' key.
+    model_config = ConfigDict(extra='allow', populate_by_name=True) 
+
+    format: str | None = None
+    datefmt: str | None = None
+    style: Literal['%', '{', '$'] | None = None
+    class_name: str | None = Field(None, alias='class')
```
tomschr: Could you add a `description` key here?


```suggestion
    class_name: str | None = Field(None, alias='class',
                                   description='The fully qualified name of the handler class',
                                   ) 
```

src/docbuild/models/config_model/app.py
```diff
@@ -1,79 +1,140 @@
...
-class App_LoggingConfig(BaseModel):
-    """Configuration for the application's logging system.
+# Ensure type annotations work correctly even if type aliases are not natively available
+if TYPE_CHECKING:
+    from typing import TypeAlias
+else:
+    TypeAlias = Any
+
+
+# --- 1. Logging Sub-Models (Schema for logging.dictConfig) ---
```
tomschr: Qualified name is incomplete:

```suggestion
# --- 1. Logging Sub-Models (Schema for logging.config.dictConfig) ---
```

src/docbuild/models/config_model/app.py
```diff
@@ -1,79 +1,140 @@
...
-    model_config = ConfigDict(extra='allow')
+    # This converts the nested Pydantic models into dictionaries using aliases ('class' instead of 'class_name')
+    # This is essential for the output to be compatible with logging.config.dictConfig.
+    @model_validator(mode='after')
+    def _export_config_as_dict(self) -> 'App_LoggingConfig':
+        """Ensure default log config is fully structured."""
+        # For simplicity in the app logic, we will assume the default dict is complete
+        # and rely on the calling function (setup_logging) to call model_dump().
+        return self
+
```
tomschr: I see still a problem with cross references. Imagine a TOML config like this:

```toml
[logging.handlers.console]
class = "logging.StreamHandler"
formatter = "simple_typo" # <--- That formatter doesn't exist

[logging.loggers.my_app]
handlers = ["console", "file_typo"] # <--- That handler doesn't exist
```

Python's logging module wouldn't work.


```suggestion
    @model_validator(mode='after')
    def _validate_cross_references(self) -> Self:
        """Checks if all references handlers and formatters exists."""
        
        # 1. Check handler references
        handler_keys = set(self.handlers.keys())
        all_logger_configs = list(self.loggers.values()) + [self.root]

        for logger_cfg in all_logger_configs:
            if logger_cfg.handlers:
                for handler_name in logger_cfg.handlers:
                    if handler_name not in handler_keys:
                        raise ValueError(
                            f"Logger {logger_cfg!r} references an unknown "
                            f"handler {handler_name!r}"
                        )

        # 2. Checks formatter references
        formatter_keys = set(self.formatters.keys())
        for handler_name, handler_cfg in self.handlers.items():
            if handler_cfg.formatter:
                if handler_cfg.formatter not in formatter_keys:
                    raise ValueError(
                        f"Handler {handler_name!r} references an unknown "
                        f"formatter {handler_cfg.formatter!r}"
                    )
        
        return self
```

tests/cli/test_cmd_cli.py
tomschr: I think we should add a new test file `tests/models/config_model/test_app.py` to collect all tests related to `models/config_model/app.py`. That new test file should contain tests related to validating the model.

When I set up the test suite, I tried to make 1:1 relationships: each Python file has an equally named file name in the tests directory with the same directory (additionally with the thed `test_` prefix required by pytest).

The file `test_cmd_cli.py` deals only with CLI parsing. IMHO it shouldn't contain other tests unrelated to CLI parsing.

src/docbuild/models/config_model/app.py
```diff
@@ -1,79 +1,140 @@
...
+    The complete Pydantic model for Python's logging.dictConfig.
+    """
+    version: Literal[1] = Field(description="The schema version. Must be 1.")
     disable_existing_loggers: bool = False
+
+    formatters: FormattersDict = Field(default_factory=dict, description="All configured formatters.")
+    filters: FiltersDict = Field(default_factory=dict, description="All configured filters.")
+    handlers: HandlersDict = Field(default_factory=dict, description="All configured handlers.")
+    loggers: LoggersDict = Field(default_factory=dict, description="All specific loggers (e.g., 'my_app').")
+    root: RootLoggerConfig = Field(default_factory=RootLoggerConfig, description="The root logger configuration.")
```
tomschr: I stumbled upon why `default_factory` is dict for the others and `RootLoggerConfig` for root. I think we should add a comment:


```suggestion
    # Use an instance as we have only one root logger
    root: RootLoggerConfig = Field(default_factory=RootLoggerConfig, description="The root logger configuration.")
```

src/docbuild/logging.py
```diff
@@ -67,15 +66,7 @@
...
-    """Create the base log directory if it doesn't exist.
-
-    This directory is typically located at :file:`~/.local/state/docbuild/logs`
-    as per the XDG Base Directory Specification.
-
-    :param base_log_dir: The base directory where logs should be stored.
-        Considers the `XDG_STATE_HOME` environment variable if set.
-    :return: The path to the base log directory.
-    """
+    """Create the base log directory if it doesn't exist."""
```
tomschr: What's wrong with the parameter descriptions? 😉 Please re-add them or change it if we need to adapt them:


```suggestion
    """Create the base log directory if it doesn't exist.
    
    This directory is typically located at :file:`~/.local/state/docbuild/logs`		
    as per the XDG Base Directory Specification.

    :param base_log_dir: The base directory where logs should be stored.
        Considers the `XDG_STATE_HOME` environment variable if set.
    :return: The path to the base log directory.
    """
```

sushant-suse: > What's wrong with the parameter descriptions? 😉 Please re-add them or change it if we need to adapt them:

Mistakenly removed it. Added it back.

tests/models/config_model/test_app.py
```diff
@@ -0,0 +1,118 @@
...
+
+
+# --- Helper Fixtures ---
+
+# The test now imports Mock and uses the function name directly
+@pytest.fixture
+def mock_replace_placeholders(monkeypatch):
+    """Mocks the replace_placeholders utility to confirm it was called."""
+    mock = Mock(wraps=lambda d: d)  # Retains original behavior but tracks calls
+    monkeypatch.setattr('docbuild.models.config_model.app.replace_placeholders', mock)
```
tomschr: This is technically correct, but there is another syntax which passes the real object. I prefer the latter for one reason: if you rename the objects, it is detected by an editor (e.g. VSCode). If the argument for `setattr` is a string, it's not as straightforward.

Something like this (untested):

```suggestion
    monkeypatch.setattr(replace_placeholders, mock)
```

tomschr:
@sushant-suse I get a strange error about locks. See the [current CI](https://github.com/openSUSE/docbuild/actions/runs/18837447609/job/53741632246?pr=91). I get this for both MacOS and Linux:

```
================== 532 passed, 3 skipped, 1 warning in 5.66s ===================
--- Logging error ---
--- Logging error ---
  File "/__t/Python/3.12.11/x64/lib/python3.12/logging/__init__.py", line 1163, in emit
    stream.write(msg + self.terminator)
ValueError: I/O operation on closed file.
  File "/__t/Python/3.12.11/x64/lib/python3.12/logging/__init__.py", line 1163, in emit
    stream.write(msg + self.terminator)
ValueError: I/O operation on closed file.
Call stack:
Call stack:
  File "/__t/Python/3.12.11/x64/lib/python3.12/threading.py", line 1032, in _bootstrap
    self._bootstrap_inner()
  File "/__t/Python/3.12.11/x64/lib/python3.12/threading.py", line 1075, in _bootstrap_inner
    self.run()
  File "/__t/Python/3.12.11/x64/lib/python3.12/threading.py", line 1012, in run
    self._target(*self._args, **self._kwargs)
  File "/__t/Python/3.12.11/x64/lib/python3.12/threading.py", line 1032, in _bootstrap
    self._bootstrap_inner()
  File "/__t/Python/3.12.11/x64/lib/python3.12/threading.py", line 1075, in _bootstrap_inner
    self.run()
  File "/__t/Python/3.12.11/x64/lib/python3.12/threading.py", line 1012, in run
    self._target(*self._args, **self._kwargs)
Message: 'Released lock at /tmp/pytest-of-root/pytest-0/test_stale_lock_is_cleaned_up0/locks/docbuild-398a3462bdef017251530b1deb3dcdae4fb72c4b59ae3177cf3feaaeb908c033.pid.'
Arguments: None
Message: 'Released lock at /tmp/pytest-of-root/pytest-0/test_stale_lock_is_cleaned_up0/locks/docbuild-398a3462bdef017251530b1deb3dcdae4fb72c4b59ae3177cf3feaaeb908c033.pid.'
Arguments: None
```

I also see this error locally when I run `upytest`. I guess, there is something fishy. :slightly_smiling_face: 

I also saw that the file `src/docbuild/utils/pidlock.py` is only ~85% covered by tests.

tomschr (approved):
Found the previous comment is unrelated to this one. Created a separate issue.

I think you can squash-merge this one and work maybe on the error?

PR #93, opened 2025-10-28
Fix #92: Correct regression on "Released lock at"

### LINKED ISSUE #92: Found "Released lock at" in CI output
### Project Version

latest

### Python Version

3.13, 3.12, 3.14

### Operating System

Linux, macOS

### Steps to Reproduce

I get a strange error about locks. See the [current CI](https://github.com/openSUSE/docbuild/actions/runs/18837447609/job/53741632246?pr=91). I get this for both MacOS and Linux:

```
================== 532 passed, 3 skipped, 1 warning in 5.66s ===================
--- Logging error ---
--- Logging error ---
  File "/__t/Python/3.12.11/x64/lib/python3.12/logging/__init__.py", line 1163, in emit
    stream.write(msg + self.terminator)
ValueError: I/O operation on closed file.
  File "/__t/Python/3.12.11/x64/lib/python3.12/logging/__init__.py", line 1163, in emit
    stream.write(msg + self.terminator)
ValueError: I/O operation on closed file.
Call stack:
Call stack:
  File "/__t/Python/3.12.11/x64/lib/python3.12/threading.py", line 1032, in _bootstrap
    self._bootstrap_inner()
  File "/__t/Python/3.12.11/x64/lib/python3.12/threading.py", line 1075, in _bootstrap_inner
    self.run()
  File "/__t/Python/3.12.11/x64/lib/python3.12/threading.py", line 1012, in run
    self._target(*self._args, **self._kwargs)
  File "/__t/Python/3.12.11/x64/lib/python3.12/threading.py", line 1032, in _bootstrap
    self._bootstrap_inner()
  File "/__t/Python/3.12.11/x64/lib/python3.12/threading.py", line 1075, in _bootstrap_inner
    self.run()
  File "/__t/Python/3.12.11/x64/lib/python3.12/threading.py", line 1012, in run
    self._target(*self._args, **self._kwargs)
Message: 'Released lock at /tmp/pytest-of-root/pytest-0/test_stale_lock_is_cleaned_up0/locks/docbuild-398a3462bdef017251530b1deb3dcdae4fb72c4b59ae3177cf3feaaeb908c033.pid.'
Arguments: None
Message: 'Released lock at /tmp/pytest-of-root/pytest-0/test_stale_lock_is_cleaned_up0/locks/docbuild-398a3462bdef017251530b1deb3dcdae4fb72c4b59ae3177cf3feaaeb908c033.pid.'
Arguments: None
```

I also see this error locally when I run `upytest`. I guess, there is something fishy. :slightly_smiling_face: 

I also saw that the file `src/docbuild/utils/pidlock.py` is only ~85% covered by tests.

_Originally posted by @tomschr in https://github.com/openSUSE/docbuild/issues/91#issuecomment-3451586666_

### Expected Behavior

_No response_

### Actual Behavior

The tests work, but we still get this error without a failing CI. This shouldn't be the case.

### Additional Information

See the CI log

I asked Gemini and it responded with this:

> First, the `PidFileLock.release` method in `src/docbuild/utils/pidlock.py` can be called during interpreter shutdown via `atexit`. At that point, the logging system may already be partially shut down, causing the log.debug call to fail with a `ValueError`: I/O operation on closed file.
>
> Second, the test `test_stale_lock_is_cleaned_up` in `tests/utils/test_pidlock.py` acquires a lock but never explicitly releases it. It relies on the `atexit` handler for cleanup, which triggers the aforementioned logging error during test execution.

tomschr:
* `PidFileLock.release` 
  The method in `src/docbuild/utils/pidlock.py` can be called during interpreter shutdown via atexit. At that point, the logging system may already be partially shut down, causing the `log.debug` call to fail with a `ValueError: I/O operation on closed file`.
* `test_stale_lock_is_cleaned_up`
  In `tests/utils/test_pidlock.py` acquires a lock but never explicitly releases it. It relies on the atexit handler for cleanup, which triggers the aforementioned logging error during test execution.

### Observations
* I don't get the same messages locally and in the CI output.

### DIFF
--- a/changelog.d/93.bugfix.rst
+++ b/changelog.d/93.bugfix.rst
@@ -1,0 +1,1 @@
+Fix regression in lock files


src/docbuild/utils/pidlock.py
```diff
@@ -5,165 +5,169 @@
...
+    def emit(self, record):
+        try:
+            super().emit(record)
+        except ValueError as e:
+            if "I/O operation on closed file" in str(e):
+                # Ignore harmless teardown errors (common in Python 3.13 + pytest)
+                return
+            raise
+
+
```
tomschr: You have a similar class in `conftest.py`. Although the general idea is a good think, "poluting" the code with test related parts should be avoided. Can we integrate that in `conftest.py`?

src/docbuild/utils/pidlock.py
```diff
@@ -5,165 +5,169 @@
...
+log.propagate = True  # Allow pytest's caplog to capture messages
+
+# Apply safe handler in CI or pytest only
+if os.getenv("CI") or os.getenv("PYTEST_CURRENT_TEST"):
+    for h in list(log.handlers):
+        log.removeHandler(h)
+    _safe_handler = _SafeStreamHandler()
+    _safe_handler.setFormatter(logging.Formatter("%(levelname)s: %(message)s"))
+    log.addHandler(_safe_handler)
+    log.propagate = False
```
tomschr: IMHO, this should also go into `conftest.py`

src/docbuild/utils/pidlock.py
```diff
@@ -5,165 +5,169 @@
...
                 self._lock_acquired = False
+                self.lock_file = None
                 log.debug("Released lock at %s.", self.lock_path)
-                
-                # Unregister the cleanup function
-                atexit.unregister(self.release)
+                try:
+                    atexit.unregister(self.release)
+                except (ValueError, AttributeError):
+                    log.debug("atexit.unregister failed or unavailable.")
```
tomschr: Maybe the order is incorrect? Shouldn't we _first_ unregister the cleanup function, and _then_ remove the file? 🤔 


```suggestion
                try:
                    atexit.unregister(self.release)
                # I'm not sure about catching these:
                except (ValueError, AttributeError):
                    log.debug("atexit.unregister failed or unavailable.")
                self.lock_path.unlink()
                self._lock_acquired = False
                self.lock_file = None
                log.debug("Released lock at %s.", self.lock_path)
```

tests/utils/test_pidlock.py
```diff
@@ -1,164 +1,181 @@
...
-def lock_setup(tmp_path: Path) -> tuple[Path, Path]:
-    """Provides a resource path and the temporary lock directory."""
-    resource_path = tmp_path / "env.production.toml"
-    resource_path.touch()  # The resource path must exist
-    lock_dir = tmp_path / "locks"
-    return resource_path, lock_dir
+    # Acquire lock
+    lock.acquire()
+    assert lock.lock_path.exists()
+    assert lock._lock_acquired
```
tomschr: I think it's safer to do that in a `with` block:


```suggestion
    with lock:
        assert lock.lock_path.exists()
        assert lock._lock_acquired
```

src/docbuild/utils/pidlock.py
```diff
@@ -5,165 +5,169 @@
...
+                    f.write(f"{current_pid}\n")
+                self._lock_acquired = True
+                self.lock_file = self.lock_path
+                log.debug(f"Acquired lock for {self.resource_path} at PID {current_pid}.")
+                atexit.register(self.release)
+                return
+            except FileExistsError:
+                pass
+            except OSError as e:
+                raise RuntimeError(f"Failed to create lock file at {self.lock_path}: {e}")
```
tomschr: I think the error message misses the point. The lock is just a technical detail. For _users_ it's important to let them know that the same instance is already running and what to do.

The following is just a suggestion, feel free to adapt it:


```suggestion
                raise RuntimeError(
                    "Same instance is already running. "
                    "Either wait until the first instance is complete or "
                    "cancel it. "
                    f"PID={current_pid}, lock file at {self.lock_path}: {e}"
                    )
```

tests/conftest.py
```diff
@@ -95,21 +96,22 @@ def pytest_configure(config):
...
-            
+                try:
+                    record = _log_queue.get_nowait()
+                    _log_listener.handle(record)
+                except queue.Empty:
+                    break
             _log_listener.stop()
         except Exception:
-            # Suppress any errors caused by closed handlers during final stop
             pass
```
tomschr: This might be quite dangerous. It can completely hide valid exceptions that should propagate the exception chain.

Can we a bit more specific about what we need to catch here?

src/docbuild/utils/pidlock.py
```diff
@@ -1,169 +1,165 @@
...
+                        except PermissionError:
+                            log.error("Non-critical error while checking lock file: permission denied")
+                        except OSError as e:
+                            log.error(f"Non-critical error while checking lock file: {e}")
+                    except ValueError:
+                        log.error(f"Invalid PID in lock file {self._lock_path}, removing")
+                        try:
+                            self._lock_path.unlink()
+                        except OSError:
+                            pass
```
tomschr: The above lines may look nicer if you use `context.suppress` like this:

```python
from contextlib import suppress
# ...

with suppress(OSError):
    self._lock_path.unlink()
```

src/docbuild/utils/pidlock.py
```diff
@@ -1,169 +1,165 @@
...
+            except OSError as e:
+                log.error(f"Non-critical error while reading lock file: {e}")
+                try:
+                    self._lock_path.unlink()
+                except OSError:
+                    pass
```
tomschr:
```suggestion
                with suppress(OSError):
                    self._lock_path.unlink()
```

src/docbuild/utils/pidlock.py
```diff
@@ -1,169 +1,151 @@
...
-import logging
-import os
 import errno
+import fcntl
+import io
+import logging
 import os
 from pathlib import Path
-from typing import Self, Any
+from typing import ClassVar, Dict, Self, cast
```
tomschr: You don't need `Dict` as we are using Python 3.12+


```suggestion
from typing import ClassVar, Self, cast
```

PR #99, opened 2025-11-17
Fix #96: Centralize Git operations

### LINKED ISSUE #96: Revisit git operations
### Situation

Currently, the code contains calls to `git clone` to create a bare Git repository. Usually it's a two steps process:

* We create a _permanent_ bare Git repository and update that whenever we access it.
* We create a _temporary_ Git repository when we need to access the content.

Not sure if accessing the Git repository with the normal Git commands is the way to go. 🤔

### Use Case

Having a Git repository is an essential part of the script on different locations:

* When creating metadata
* When initiating a build

### Possible Implementation

Dealing with the Git repo through a shell command may create another layer which may be inefficient.

Perhaps we should think about using one of the native libraries/Python bindings to Git.

tomschr:
Refactor some git related commands.

# Reason
I discovered that some `git` related shell code was repeated in different files, so it was time to consolidate it and collect it under one roof.

# Implementation

Introduce:

* a general `run_command` which can run any command.
* a specific `execute_git_command` for general git commands.
* a `shell.py` module and collects the previously mentioned functions.
* a `ManagedGitRepo` class that manages bare repos and its worktrees.

## Relationship

Neded for #4

### DIFF
--- a/changelog.d/99.feature.rst
+++ b/changelog.d/99.feature.rst
@@ -1,0 +1,1 @@
+Centralize Git and shell operations for better code reuse across modules


sushant-suse:
Hi @tomschr, thanks for the addition. This looks great to me. I just have a few minor questions. Maybe that can help?

In the issue, we wanted to create a permanent bare Git repository and update that whenever we access it. The current `ManagedGitRepo` class provides:
- `clone_bare()`: for creation
- `create_worktree()` (or `clone_local_copy`): for the temporary access

IMO, there should also be a method for updating the permanent bare repository, which is typically done with the `git fetch` command.

We can implement a method for this. 
<details>
  <summary>Under the `ManagedGitRepo` (in src/docbuild/utils/git.py):</summary>

```python
async def fetch_updates(self) -> bool:
    """Fetch updates from the remote to the bare repository."""
    if not self.bare_repo_path.exists():
        log.warning(
            'Cannot fetch updates: Bare repository does not exist at %s',
            self.bare_repo_path
        )
        return False
        
    log.info("Fetching updates for '%s'", self.slug)
    try:
        self.stdout, self.stderr = await execute_git_command(
            'fetch', 
            '--all', 
            cwd=self.bare_repo_path
        )
        log.info("Successfully fetched updates for '%s'", self.slug)
        return True
    except RuntimeError as e:
        log.error("Failed to fetch updates for '%s': %s", self.slug, e)
        return False
```

</details>

I think we will also need to add a corresponding test `test_managed_repo_fetch_updates_success` in the `tests/utils/test_git.py` file.

WDYT?

src/docbuild/utils/shell.py
```diff
@@ -0,0 +1,87 @@
...
+    # Default Git arguments for consistent behavior
+    default_git_args = ('-c', 'color.ui=never')
+    command = ('git', *default_git_args, *args)
+    log.debug('Executing Git command: %s in %s', ' '.join(command), cwd)
+
+    # Default environment for secure and non-interactive execution
+    default_env = {
+        'LANG': 'C',
+        'LC_ALL': 'C',
+        'GIT_TERMINAL_PROMPT': '0',
```
sushant-suse: In the old `clone_repo` function (under the file `src/docbuild/cli/cmd_repo/process.py`), we had an environment variable - `GIT_PROGRESS_FORCE: '1'`. This is important for capturing progress output. 

You fixed this in by adding it to the default environment in the current function (`execute_git_command`). Should we also add that variable here like?
```suggestion
        'GIT_TERMINAL_PROMPT': '0',
        'GIT_PROGRESS_FORCE': '1',
```

sushant-suse:
> Hi @tomschr, thanks for the addition. This looks great to me. I just have a few minor questions. Maybe that can help?
> 
> In the issue, we wanted to create a permanent bare Git repository and update that whenever we access it. The current `ManagedGitRepo` class provides:
> 
> * `clone_bare()`: for creation
> * `create_worktree()` (or `clone_local_copy`): for the temporary access
> 
> IMO, there should also be a method for updating the permanent bare repository, which is typically done with the `git fetch` command.
> 
> We can implement a method for this.
> 
> Under the `ManagedGitRepo` (in src/docbuild/utils/git.py):
> ```python
> async def fetch_updates(self) -> bool:
>     """Fetch updates from the remote to the bare repository."""
>     if not self.bare_repo_path.exists():
>         log.warning(
>             'Cannot fetch updates: Bare repository does not exist at %s',
>             self.bare_repo_path
>         )
>         return False
>         
>     log.info("Fetching updates for '%s'", self.slug)
>     try:
>         self.stdout, self.stderr = await execute_git_command(
>             'fetch', 
>             '--all', 
>             cwd=self.bare_repo_path
>         )
>         log.info("Successfully fetched updates for '%s'", self.slug)
>         return True
>     except RuntimeError as e:
>         log.error("Failed to fetch updates for '%s': %s", self.slug, e)
>         return False
> ```
> 
> I think we will also need to add a corresponding test `test_managed_repo_fetch_updates_success` in the `tests/utils/test_git.py` file.
> 
> WDYT?

One more thing that came to my mind (if this one makes sense then only😂) - 

We will also need tot update the `process` logic iIntegration. Once `fetch_updates` is implemented, the main logic in the `src/docbuild/cli/cmd_repo/process.py` should also be updated to use it.

The original logic is:
```python
if repo_path.exists():
    log.info('Repository %r already exists...')
    return True
```

The new logic should look like (just a rough idea):
```python
if repo.bare_repo_path.exists():
    # If it exists, UPDATE it before proceeding
    log.info('Repository already exists, fetching updates...')
    # Now calls the new method
    return await repo.fetch_updates()
else:
    # If it does not exist, CLONE it
    return await repo.clone_bare()
```

src/docbuild/utils/shell.py
tomschr: Yes, you are right. Good catch! :+1:

tomschr:
Thank you for all the comments. :+1:

There is one thing in your https://github.com/openSUSE/docbuild/pull/99#issuecomment-3545091930. You suggested to fetch updates when the clone already exists. In theory this sounds nice. But it could end up soon in the following scenario:

1. Script starts.
2. The repo got updated. All is fine.
3. While the script is running, in the meantime another update is pushed to the repo.
4. When the script hits the clone_bare method again, it will update the repo.
5. Boom, there are different states.

So we need to make sure to update this particular repo only once. One way to do it is something like this:

```python
class ManagedGitRepo:
    """Manages a bare repository and its temporary worktrees."""
    #: Class variable to indicate the update state of a repo
    _is_updated: ClassVar[dict[Repo, bool]] = {}

    def __init__(self: Self, remote_url: str, permanent_root: Path) -> None:
       # ...
       # Add repo into
       type(self)._is_updated[self._repo_model] = False

    async def clone_bare(self: Self) -> bool:
       # ...
       cls = type(self)
       if self.bare_repo_path.exists() and not cls._is_updated.get(self._repo_model, False):
            log.info('Repository already exists at %s', self.bare_repo_path)
            cls._is_updated[self._repo_model] = True
            return await self.fetch_updates()
```

What do you think?

sushant-suse:
> Thank you for all the comments. 👍
> 
> There is one thing in your [#99 (comment)](https://github.com/openSUSE/docbuild/pull/99#issuecomment-3545091930). You suggested to fetch updates when the clone already exists. In theory this sounds nice. But it could end up soon in the following scenario:
> 
> 1. Script starts.
> 2. The repo got updated. All is fine.
> 3. While the script is running, in the meantime another update is pushed to the repo.
> 4. When the script hits the clone_bare method again, it will update the repo.
> 5. Boom, there are different states.
> 
> So we need to make sure to update this particular repo only once. One way to do it is something like this:
> 
> ```python
> class ManagedGitRepo:
>     """Manages a bare repository and its temporary worktrees."""
>     #: Class variable to indicate the update state of a repo
>     _is_updated: ClassVar[dict[Repo, bool]] = {}
> 
>     def __init__(self: Self, remote_url: str, permanent_root: Path) -> None:
>        # ...
>        # Add repo into
>        type(self)._is_updated[self._repo_model] = False
> 
>     async def clone_bare(self: Self) -> bool:
>        # ...
>        cls = type(self)
>        if self.bare_repo_path.exists() and not cls._is_updated.get(self._repo_model, False):
>             log.info('Repository already exists at %s', self.bare_repo_path)
>             cls._is_updated[self._repo_model] = True
>             return await self.fetch_updates()
> ```
> 
> What do you think?

Great observation @tomschr. This is a critical point regarding consistency. Fetching updates on subsequent calls within the same execution cycle could lead to state discrepancies, especially if the underlying data is being used for different metadata creation steps. The idea of using a class-level cache (`_is_updated: ClassVar[dict[Repo, bool]]`) is a standard way to implement idempotent operations within a single process run.

I think, the logic flow looks slightly cleaner if we handle the "already updated" state first and then proceed with the "needs update/clone" logic. Also, using the boolean result of the check directly simplifies the if statement?

Something like:
```python
class ManagedGitRepo:
    """Manages a bare repository and its temporary worktrees."""

    _is_updated: ClassVar[dict['Repo', bool]] = {} 

    def __init__(self: Self, remote_url: str, permanent_root: Path) -> None:
       # ...
       self._repo_model = Repo(remote_url)

       if self._repo_model not in type(self)._is_updated:
           type(self)._is_updated[self._repo_model] = False

    async def clone_bare(self: Self) -> bool:
        cls = type(self)
        
        if cls._is_updated.get(self._repo_model, False):
            log.info('Repository %r already processed this run.', self._repo_model.name)
            return True # Successfully short-circuit

        if self.bare_repo_path.exists():
            log.info('Repository already exists, fetching updates for %r.', self._repo_model.name)
            success = await self.fetch_updates()
        else:
            log.info('Repository does not exist, cloning bare for %r.', self._repo_model.name)
            success = await self._perform_initial_clone() # Assuming you extract clone logic here for clarity

        if success:
            cls._is_updated[self._repo_model] = True
        
        return success
```

PR #101, opened 2025-11-19
Fix #21: fixed the use of Pydantic for validation of env config

### LINKED ISSUE #21: Use Pydantic for validation of env config
### Situation

Currently, the configuration for the docbuild _environment_ isn't validated against errors.
The issue about docbuild app config is described in #88.

### Use Case

Having a mechanism for validating the content of the env config helps in discovering configuration problems like typos or improper types.

### Possible Implementation

Use Pydantic and create a hierarchy of classes. Each class serves as a "section" in the TOML file. It's properties are the respective keys with their values (see below as an example).

* Prefer `Field()` for each attribute.
* Use `title`, `description`, and possible `examples` as arguments in `Field`. They are used in the documentation.
* Add an additional docstring _after_ `Field()` (yes, it's a weird requirement for our documentation build tool)

<details>
  <summary>Draft of the env config model</summary>

```python
"""Pydantic models for application and environment configuration."""

from pathlib import Path
from typing import Annotated

from pydantic import BaseModel, Field, HttpUrl, IPvAnyAddress

from ..models.language import LanguageCode

# A type for domain names, validated with a regex.
# This ensures the string looks like a valid domain, but doesn't do a DNS lookup.
# TODO: Add title, description, examples
DomainName = Annotated[
    str,
    Field(
        pattern=r"^(?:[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?\.)+[a-zA-Z]{2,63}$"
    ),
]


class Env_Server(BaseModel):
    """Defines server settings."""

    # TODO: Add Field and title, description, examples
    name: str
    role: str
    host: IPvAnyAddress | DomainName
    port: int | None = None
    enable_mail: bool


class Env_GeneralConfig(BaseModel):
    """Defines general configuration."""

    # TODO: Add Field and title, description, examples
    default_lang: LanguageCode
    languages: list[LanguageCode]
    canonical_url_domain: HttpUrl


class Env_TmpPaths(BaseModel):
    """Defines temporary paths."""

    # TODO: Add Field and title, description, examples
    tmp_base_path: Path
    tmp_path: Path
    tmp_deliverable_path: Path
    tmp_build_dir: Path
    tmp_out_path: Path
    log_path: Path
    tmp_deliverable_name: str


class Env_TargetPaths(BaseModel):
    """Defines target paths."""

    # TODO: Add Field and title, description, examples
    target_path: str
    backup_path: Path


class Env_PathsConfig(BaseModel):
    """Defines various application paths."""

    # TODO: Add Field and title, description, examples
    config_dir: Path
    repo_dir: Path
    temp_repo_dir: Path
    base_cache_dir: Path
    cache_dir: Path
    meta_cache_dir: Path
    tmp: TmpPaths
    target: Env_TargetPaths


class EnvConfig(BaseModel):
    """Root model for the environment configuration (env.toml)."""

    # TODO: Add Field and title, description, examples
    server: Env_Server
    config: Env_GeneralConfig
    paths: Env_PathsConfig
    xslt_params: dict[str, str | int] = Field(alias='xslt-params')
```

Additionally, we need to rethink how we want to deal with the placeholders. This can be done in two ways:

1. Load the raw data, replace the placeholders, and pass the structure to `EnvConfig`.
2. Let Pydantic's method `model_validator` deal with the replacement *inside*.

To make it possible, you need to add in `EnvConfig`:

```python
from copy import deepcopy
from pydantic import BaseModel, Field, HttpUrl, IPvAnyAddress, model_validator

class EnvConfig(BaseModel):
    # ...
    @model_validator(mode='before')
    @classmethod
    def _resolve_placeholders(cls, data: Any) -> Any:
        """Resolve placeholders before any other validation."""
        if isinstance(data, dict):
            # Make a deep copy to avoid modifying the original input data,
            # as replace_placeholders works in-place.
            return replace_placeholders(deepcopy(data))
        return data
```
</details>

The execution flow looks like this:

1. You call `EnvConfig.model_validate(raw_config_dict)`.
1. Pydantic sees the `@model_validator(mode='before')` on `EnvConfig`.
1. It calls our `_resolve_placeholders` method, passing it the `raw_config_dict`.
1. Our method calls `replace_placeholders(raw_config_dict)`, which walks the entire dictionary and resolves all placeholders **in one pass**.
1. Pydantic then takes the processed dictionary (with all placeholders resolved) and begins its normal validation, creating instances of `Env_ServerConfig` etc., from the clean data.

sushant-suse:
Related Issue: #21 

#### 1. `src/docbuild/models/config_model/env.py`

* **Changed Made**:  Defines the complete `EnvConfig` Pydantic hierarchy (`Env_Server`, `Env_PathsConfig`, etc.).
* **Why Changed**: This file creates the definitive schema for `env.toml`. It enforces strict type checking (e.g., ensuring paths are `Path` objects and hostnames are valid domains/IPs) and includes the `@model_validator(mode='before')` hook to process and resolve all placeholders (`{...}`) in the raw dictionary before Pydantic begins validation.

#### 2. `src/docbuild/models/config_model/__init__.py`

* **Changed Made**: Added export statements for the new environment model.
* **Why Changed**: Exposes the new model (`EnvConfig`) to the application package, allowing other modules (like `cmd_cli.py`) to import it using relative paths (`from ..models.config_model.env import EnvConfig`).

#### 3. `tests/models/config_model/__init__.py`

* **Changed Made**: Standard Python packaging requirement for the test directory structure.
* **Why Changed**: Ensures the test structure mirrors the source structure, allowing test frameworks to correctly discover unit tests within the `tests/models/config_model` directory.

#### 4. `src/docbuild/cli/cmd_cli.py`

* **Changed Made**: Updated Phase 2 (Environment Config Load) to import and use `EnvConfig.from_dict(raw_env_data)`. The manual call to `replace_placeholders(...)` was removed.
* **Why Changed**: This completes the integration. The raw environment data is now validated against the new schema, and the validated Pydantic object replaces the raw dictionary in the context. The removal of the manual placeholder call shifts that responsibility entirely to the Pydantic model's `model_validator` hook.

#### 5. `src/docbuild/config/load.py`

* **Changed Made**: Removed the obsolete function `process_envconfig()` and its associated logic.
* **Why Changed** :The logic previously housed in `process_envconfig` (file searching and raw loading) is now fully handled by the generic `handle_config` function and the subsequent Pydantic validation in the CLI. Removing the dedicated, now-redundant function cleans up the core library.

#### 6. `tests/cli/cmd_config/test_config.py`

* **Changed Made:**: Modified the `mock_full_cli_setup` fixture to provide two sequential return values for `handle_config` (AppConfig data first, then EnvConfig data).
* **Why Changed**: This sequence mock is crucial to prevent the CLI from crashing (`SystemExit(1)`) during initialization, as the main `cli()` function calls the config loader twice in a mandatory order.

#### 7. `tests/cli/test_cmd_cli.py`

* **Changed Made**: Updated the primary `mock_config_models` fixture to ensure mock `AppConfig` instances correctly expose the `.logging.model_dump` method (to satisfy logging setup). Also, updated test assertions to expect Pydantic objects instead of raw dictionaries.
* **Why Changed**: Guarantees that the integration tests correctly verify the full application lifecycle, from raw data loading to validated object storage in the `DocBuildContext`.

#### 8. `tests/config/test_load.py`

* **Changed Made**: Deleted all tests that called the now-removed `process_envconfig` function.
* **Why Changed**: This resolves immediate test failures and ensures the file only tests the generic `handle_config` utility, aligning the test file with the simplified source code (`src/docbuild/config/load.py`).

#### 9. `tests/conftest.py`

* **Changed Made**: Removed obsolete fixtures (like `fake_envfile`) that were trying to patch the deleted function `process_envconfig`.
* **Why Changed**: Removes residual dependency patching that would otherwise cause an `AttributeError` during test collection and setup.

#### 10. `tests/cli/cmd_config/test_environment.py`

* **Changed Made**: Deleted the file.
* **Why Changed**: This test file was causing a fatal `AttributeError` during test setup because its fixtures relied on the obsolete `process_envconfig` function. Its removal stabilizes the overall test suite.


src/docbuild/models/config_model/env.py
```diff
@@ -0,0 +1,268 @@
...
+
+    canonical_url_domain: HttpUrl = Field(
+        title="Canonical URL Domain",
+        description="The base domain used to construct canonical URLs for SEO purposes.",
+        examples=["https://docs.example.com"],
+    )
+    "The canonical domain for URLs."
+
+
+class Env_TmpPaths(BaseModel):
```
tomschr: After looking over the names, I think I've introduced some, aehm, inconsistencies.For example, we have `tmp_base_path` (suffix "path") and `tmp_build_dir` (suffix "dir").

I think it would be a good timing to fix that as well. Sorry for the inconvenience! I think there are a few locations where these names are used.

Additionally, you could also adapt `etc/docbuild/env.example.toml` too as this acts as a template.

src/docbuild/models/config_model/env.py
```diff
@@ -0,0 +1,268 @@
...
+
+    temp_repo_dir: Path = Field(
+        title="Temporary Repository Directory",
+        description="The directory used for temporary working copies cloned from the permanent bare repositories.",
+    )
+    "Path for temporary working copies."
+
+    base_cache_dir: Path = Field(
+        title="Base Cache Directory",
+        description="The root directory for all application-level caches.",
```
tomschr:
```suggestion
        description="The root directory for all application-level caches.",
        examples=["/var/cache/docserv"],
```

src/docbuild/models/config_model/env.py
```diff
@@ -0,0 +1,268 @@
...
+        title="Path Configuration",
+        description="All file system path definitions.",
+    )
+    "File system paths."
+
+    xslt_params: dict[str, str | int] = Field(
+        default_factory=dict,
+        alias='xslt-params',
+        title="XSLT Parameters",
+        description="Custom parameters passed directly to the DAPS XSLT processor.",
```
tomschr:
```suggestion
        description="Custom XSLT parameters passed directly to DAPS.",
```

tomschr (changes requested):
Found a minor thing that I forgot, sorry.

Regarding the coverage, I see:

```
src/docbuild/models/path.py                       35      3      8      1  90.7%   30, 64-65
```

Line 30 will be automatically gone if you apply the `Path(path).expanduser().resolve()` fix. Could you check the other lines and cover them? It's a check whether the path exists and the test should catch the OSError.

tomschr (changes requested):
Ok, so this time I didn't look at the code and just run the `config` subcommand:

```
docbuild  --env-config=etc/docbuild/env.example.toml config env
```

This gave me many validation errors:

<details>
   <summary>Validation errors</summary>

```
[2025-11-26 12:08:12] [ERROR] docbuild.cli.cmd_cli: Environment configuration failed validation: Error in config file(s): (PosixPath('etc/docbuild/env.example.toml'),) 19 validation errors for EnvConfig
paths.tmp.tmp_path
  Field required [type=missing, input_value={'tmp_base_dir': '/var/tm...{docset}_{lang}_XXXXXX'}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.12/v/missing
paths.tmp.tmp_deliverable_path
  Field required [type=missing, input_value={'tmp_base_dir': '/var/tm...{docset}_{lang}_XXXXXX'}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.12/v/missing
paths.tmp.tmp_build_dir
  Value error, Could not create directory '/var/tmp/docbuild/doc-example-com/build/{product}-{docset}-{lang}': [Errno 13] Permission denied: '/var/tmp/docbuild/doc-example-com/build' [type=value_error, input_value=PosixPath('/var/tmp/docbu...oduct}-{docset}-{lang}'), input_type=PosixPath]
    For further information visit https://errors.pydantic.dev/2.12/v/value_error
paths.tmp.tmp_out_path
  Field required [type=missing, input_value={'tmp_base_dir': '/var/tm...{docset}_{lang}_XXXXXX'}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.12/v/missing
paths.tmp.log_path
  Field required [type=missing, input_value={'tmp_base_dir': '/var/tm...{docset}_{lang}_XXXXXX'}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.12/v/missing
paths.tmp.tmp_dir
  Extra inputs are not permitted [type=extra_forbidden, input_value='/var/tmp/docbuild/doc-example-com', input_type=str]
    For further information visit https://errors.pydantic.dev/2.12/v/extra_forbidden
paths.tmp.tmp_metadata_dir
  Extra inputs are not permitted [type=extra_forbidden, input_value='/var/tmp/docbuild/doc-example-com/metadata', input_type=str]
    For further information visit https://errors.pydantic.dev/2.12/v/extra_forbidden
paths.tmp.tmp_deliverable_dir
  Extra inputs are not permitted [type=extra_forbidden, input_value='/var/tmp/docbuild/doc-example-com/deliverable/', input_type=str]
    For further information visit https://errors.pydantic.dev/2.12/v/extra_forbidden
paths.tmp.tmp_out_dir
  Extra inputs are not permitted [type=extra_forbidden, input_value='/var/tmp/docbuild/doc-example-com/out/', input_type=str]
    For further information visit https://errors.pydantic.dev/2.12/v/extra_forbidden
paths.tmp.log_dir
  Extra inputs are not permitted [type=extra_forbidden, input_value='/var/tmp/docbuild/doc-example-com/log/', input_type=str]
    For further information visit https://errors.pydantic.dev/2.12/v/extra_forbidden
paths.target.target_path
  Field required [type=missing, input_value={'target_dir': 'doc@10.10...build/external-builds/'}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.12/v/missing
paths.target.backup_path
  Field required [type=missing, input_value={'target_dir': 'doc@10.10...build/external-builds/'}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.12/v/missing
paths.target.target_dir
  Extra inputs are not permitted [type=extra_forbidden, input_value='doc@10.100.60.1:/srv/docs', input_type=str]
    For further information visit https://errors.pydantic.dev/2.12/v/extra_forbidden
paths.target.backup_dir
  Extra inputs are not permitted [type=extra_forbidden, input_value='/data/docbuild/external-builds/', input_type=str]
    For further information visit https://errors.pydantic.dev/2.12/v/extra_forbidden
paths.root_config_dir
  Extra inputs are not permitted [type=extra_forbidden, input_value='/etc/docbuild', input_type=str]
    For further information visit https://errors.pydantic.dev/2.12/v/extra_forbidden
paths.jinja_dir
  Extra inputs are not permitted [type=extra_forbidden, input_value='/etc/docbuild/jinja-doc-suse-com', input_type=str]
    For further information visit https://errors.pydantic.dev/2.12/v/extra_forbidden
paths.server_rootfiles_dir
  Extra inputs are not permitted [type=extra_forbidden, input_value='/etc/docbuild/server-root-files-doc-suse-com', input_type=str]
    For further information visit https://errors.pydantic.dev/2.12/v/extra_forbidden
paths.base_server_cache_dir
  Extra inputs are not permitted [type=extra_forbidden, input_value='/var/cache/docserv/doc-example-com', input_type=str]
    For further information visit https://errors.pydantic.dev/2.12/v/extra_forbidden
paths.base_tmp_dir
  Extra inputs are not permitted [type=extra_forbidden, input_value='/var/tmp/docbuild', input_type=str]
    For further information visit https://errors.pydantic.dev/2.12/v/extra_forbidden
```
</details>

Perhaps I should probably explain it better. 🙂 Ok, here is the thing. Inside the ENV config, we have two kind of potential placeholders:

* **Regular/config placeholders** Syntax `{placeholder}`
  They are used to reference _other parts of the ENV configuration_. For example, if you have `"{tmp_base_dir}/doc-example-com"` as string, the placeholder `tmp_base_dir` is searched in the current section of the config.
  It's possible to refer to other sections of the config with `section.key` like in `paths.base_tmp_dir`.
  Whatever variation we use, at the end it should resolve to at least some path that _can_ be accessed.

* **Runtime placeholders** Syntax `{{placeholder}}`
  These types of placeholders cannot and must not be replaced. They are replaced at runtime with specific values.

Unfortunately, the tricky part is, that sometimes we have both types in our TOML:

```toml
tmp_build_dir = "{tmp_dir}/build/{{product}}-{{docset}}-{{lang}}" 
```

You can only resolve the first placeholders, but not the others. Trying to create that directory will fail (as seen above in the error messages).

I would suggest to investigate the following possible solutions to mitigate this:

1. Maybe we introduce a better Pydantic type? A type that can deal with these placeholders?
   For examples, it could hold the resolved part and if it still sees placeholders in it, it doesn't try to resolve it. Maybe it can try to resolve _some_ parts. In the example above, we could try to resolve `{tmp_dir}` placeholder and see if it has the right privileges. But we can't do more.
   Maybe we need to rewrite and rename `EnsureWritableDirectory` to something more appropriate?

1. Add specific unit tests.
  You added `tests/models/config_model/test_env.py` which is fine, but maybe... a bit too general. All of your tests use `EnvConfig`. Perhaps it would be easier if you test our _individual_ models: a test for `Env_BuildDaps`, another for `Env_BuildContainer` etc.
  That would make the tests shorter as you don't have to deal with all sorts of secondary problems.
  One way to organize it could be to create a `tests/models/config_model/env/` directory and add individual files (like `test_env_builddaps.py`, `test_env_buildcontainer.py` etc.) for each of the models.
  Maybe rename `tests/models/config_model/test_env.py` to `tests/models/config_model/test_envconfig.py` to make it more obvious, but maybe that's just an idea.

1. Maybe we should disallow the mixture of regular and runtime placeholders in TOML config?
  That would be one way to make it cleaner and more obvious. But that would probably need changes in other parts. Or we would have to concatenate two config keys, one for the directory, another one of the last part:
  
  ```toml
  tmp_build_dir = "{tmp_dir}/build/"                # <= maybe type EnsureWritableDirectory?
  build_fragment="{{product}}-{{docset}}-{{lang}}"  # <= type str
  ```

Another thing that caught my eye:

* We have `src/docbuild/models/config_model`. I think, that's a misnomer as we have "models" already in the parent directory. Why not just `config`? Would be shorter and easier to type. 😉

Ok, thanks!

sushant-suse:
> Thanks Sushant for your reply! 👍
> 
> > [...]
> > I've pushed an update that resolves all 19 validation errors you encountered and stabilizes the configuration load process. All unit tests, including those covering configuration validation, are now passing.
> 
> Ok, I've tested it with `etc/docbuild/env.example.toml` and I don't see a difference. 🤔 I hope I didn't miss anything, but I still get 18 validation errors, before it was 19. One less!
> 
> Used command:
> 
> ```
> docbuild -vv --env-config etc/docbuild/env.example.toml config
> ```
> 
> > I appreciate the suggestions for further cleanup and test organization:
> > Placeholder Separation (Solution 3): I agree that separating paths into constant bases and runtime fragments (`tmp_build_base_dir` vs. `build_fragment`) is the cleanest long-term solution. I've deferred this change to a follow-up ticket, as implementing it now would require changes in the build logic that consumes these paths, not just in the configuration model itself.
> 
> Yes, I agree. I guess, the placeholder separation is the cleaner approach. I can create this separate ticket.
> 
> > Specific Unit Tests (Granularity): I also agree this is a strong organizational improvement. I plan to address this in a follow-up commit dedicated to refactoring `tests/models/config_model/test_env.py` into smaller, more targeted test files.
> 
> 👍 I really think that makes it easier as it only tests _one specific_ class and not the whole big beast.
> 
> > Directory Naming (models/config_model to config): Renaming major directories carries a high risk of merge conflicts and requires changes across many files. I've deferred this large structural change to a separate refactoring ticket.
> > The current changes should fulfill the functional requirements for configuration validation. Please let me know if you spot any other issues.
> 
> Yes, that can be done in a separate ticket if you feel more comfortable. I'll open a ticket for that as well.

Thanks @tomschr, I have fixed the validation error. Please have a look.
Thanks.

Also thanks for creating the tickets.

tomschr:
> I have fixed the validation error. Please have a look.

Yes, that works a lot better! :+1:  Thank you very much!

There is one thing left when I run the `docbuild config env` command:

```
$ docbuild  --env-config etc/docbuild/env.example.toml config env
# ENV Config file 'etc/docbuild/env.example.toml'
EnvConfig(
    server=Env_Server(
        name='doc-example-com',
        role=<ServerRole.PRODUCTION: 'production'>,
        host=IPv4Address('127.0.0.1'),
        enable_mail=True
    ),
[...]
```

It outputs the expected config, but we get the class names in the output like `Env_Server`, `Env_PathsConfig` etc.

I'm not sure how the the exact technical term for Pydantic is named, but don't we need to _serialize it back to string_? For example, we have the output:

```
temp_repo_dir=PosixPath('/data/docserv/repos/temporary-branches')
```

It should be:

```
temp_repo_dir='/data/docserv/repos/temporary-branches'
```

If we can do that, then I think we are ready to merge it.

src/docbuild/models/path.py
```diff
@@ -0,0 +1,92 @@
...
+        if missing_perms:
+            raise ValueError(
+                f"Insufficient permissions for directory '{path}'. "
+                f"Missing: {', '.join(missing_perms)}"
+            )
+
+        # Return an instance of the custom type (the __init__ method runs next)
+        return cls(path)
+
+    # --- Usability Methods ---
```
tomschr: I played around a bit and with this code the `EnsureAccessDirectory` output vanish:


```suggestion
    @classmethod
    def __get_pydantic_core_schema__(
        cls, source_type: Any, handler: GetCoreSchemaHandler
    ) -> core_schema.CoreSchema:
        """Define Validation AND Serialization logic."""

        # 1. Validation Logic (The Chain)
        validation_schema = core_schema.chain_schema(
            [
                handler(Path),
                core_schema.no_info_plain_validator_function(cls.validate_and_create),
            ]
        )

        # 2. Serialization Logic (Convert to String)
        # This tells Pydantic: "When dumping to JSON or Python, use str(instance)"
        serialization_schema = core_schema.plain_serializer_function_ser_schema(
            lambda instance: str(instance),
            when_used='always',  # Apply to both JSON and Python (dict) dumps
        )

        # 3. Combine Validation and Serialization
        return core_schema.json_or_python_schema(
            json_schema=validation_schema,
            python_schema=validation_schema,
            serialization=serialization_schema,
        )

    # --- Usability Methods ---
```

Not sure if this is the "right way" to do it. 😉

tomschr (changes requested):
I've played around a bit and came closer to the expected output. However, we still have a discrepancy here.

For example, the languages are written like:

```json
{
    "config": {
        "default_lang": {
            "language": "en-us",
            "lang": "en",
            "country": "us"
        },
        "languages": [
           {
                "language": "de-de",
                "lang": "de",
                "country": "de"
            }
        ]
    }
}
```

That output is wrong. It should be:

```json
{
    "config": {
        "default_lang": "en-us",
        "languages": ["de-de", "..."]
    }
}
```

I'm not sure how to avoid it. We don't need the "class fluff" in the output. 😉

PR #116, opened 2025-12-10
fix #109: Rename models/config_model to models/config

### LINKED ISSUE #109: Rename directory `models/config_model` to `models/config`
### Situation

The directory name `models/config_model` is a bit of a misnomer as the "model" is already in the `models` parent directory. As such `config` is enough.

### Use Case

* Easier to type
* Shorter
* Doesn't repeat its purpose

### Possible Implementation

Using VSCode, it may be even a simple task by renaming the directory and the editor does the rest. A simple grep should help to find all matches.

sushant-suse:
Linked Issue #109

### DIFF
--- a/src/docbuild/cli/cmd_cli.py
+++ b/src/docbuild/cli/cmd_cli.py
@@ -14,8 +14,8 @@ from typing import Any, cast
 from ..__about__ import __version__
 from ..config.app import replace_placeholders
 from ..config.load import handle_config
-from ..models.config_model.app import AppConfig
-from ..models.config_model.env import EnvConfig 
+from ..models.config.app import AppConfig
+from ..models.config.env import EnvConfig
 
 from ..constants import (
     APP_CONFIG_BASENAMES,


tomschr (changes requested):
This looks good. There is only two lines you forgot in `src/docbuild/cli/cmd_cli.py`:

```diff
diff --git i/src/docbuild/cli/cmd_cli.py w/src/docbuild/cli/cmd_cli.py
index ad98051..6299c5c 100644
--- i/src/docbuild/cli/cmd_cli.py
+++ w/src/docbuild/cli/cmd_cli.py
@@ -14,8 +14,8 @@ from typing import Any, cast
 from ..__about__ import __version__
 from ..config.app import replace_placeholders
 from ..config.load import handle_config
-from ..models.config_model.app import AppConfig
-from ..models.config_model.env import EnvConfig 
+from ..models.config.app import AppConfig
+from ..models.config.env import EnvConfig
```

If you apply these, then my tests succeeds. 😸

PR #134, opened 2026-01-13
Reduce coverage report

tomschr:
A lengthy coverage report were most files are 100% aren't easy to read. Adding the respective options in `.coveragerc` leads to a condensed form.

Additionally, the coverage report is sorted in descending order of the coverage percentage value. See coverage report below.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -7,6 +7,11 @@ omit =
 
 [report]
 precision = 1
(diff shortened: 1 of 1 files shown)


sushant-suse (approved):
Thanks Toms, this makes more sense and reduces the clutter.

Issue #110, opened 2025-12-03
Document env config


tomschr:
### Affected Documentation Section

Reference & User

### Suggested Change

When the environment config is stable enough, we need to documenting the following:

* showing the respective keys, their meaning, and their datatypes
  => This is pure referential matter and needs to go into the Reference 
* using the `config` subcommand
  => This should cover how to view the env config with all resolved fields. This should go into the User Guide. Although we have #112 which isn't finished yet, I think the change should be minimal.

### Additional Context

See related issue #21 and #112

PR #156, opened 2026-02-06
docs: add environment config reference and user guide for config subc…

sushant-suse:
Reference Issue #110

### DIFF
--- a/src/docbuild/models/config/env.py
+++ b/src/docbuild/models/config/env.py
@@ -55,9 +55,9 @@ class EnvBuildDaps(BaseModel):
 
     meta: str = Field(
         ...,
-        title="DAPS Metadata Command",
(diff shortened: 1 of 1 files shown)


src/docbuild/models/config/env.py
```diff
@@ -43,19 +43,37 @@
...
+        description="The base daps command executable.",
+        examples=["daps"]
+    )
+    "The base command used for DAPS execution."
+
+    meta: str = Field(
+        ...,
+        title="DAPS Metadata Command",
+        description="The daps metadata command for extracting info.",
+        examples=["daps-xml-meta"]
```
tomschr: That's the real subcommand:

```suggestion
        examples=["daps metadata"]
```

PR #163, opened 2026-02-10
feat(cli): add 'check files' command with optimized git grouping

### LINKED ISSUE #78: Introduce new subcommand to check availability of DC files in repos
### Situation

We all know this: when time is limited, we forget things. This could lead to a situation where the DC files in the config dont't reflect the current state in the Git repo.

### Use Case

Before we should start a build we need to check if all DC files are available. This avoid broken builds.

### Possible Implementation

Introduce a new CLI subcommand `status`, `verify` or `check` which could determine whether a DC file is available in a repo.

The potential implementation could look like:

1. Clone or update the cached repos (this code is already available).
   Use the permanent bare repo with `git --git-dir=DIR ls-tree --name-only -r BRANCH`. No need to clone the bare repo.
2. Load or access the XML config tree.
3. Iterate over all deliverables and extract the DC filename:
   a. Determine the repo and the branch of the deliverable
   b. Check if the DC filename is available in the repo.
   c. Store the result.
4. Collect all results of the check and output all files that couldn't be found. The result should contain the repo, branch, and the DC filename that isn't available.

#### Optimization

Perhaps one way to do it to group the deliverables per repo and branch and check them in one go. This avoids switching between repos and branches all the time.

#### Relationship

Maybe this needs to be grouped with #9

sushant-suse:
Fixes #78

### Files changed

* **`src/docbuild/cli/cmd_cli.py`**: Registered the new `check` command group and `files` sub-command.
* **`src/docbuild/utils/git.py`**: Enhanced `ManagedGitRepo` with `clone_bare`, `_fetch_bare`, and `ls_tree` methods to support high-performance file verification without full checkouts.
* **`src/docbuild/cli/cmd_check/`**: Created new module containing:
    * `__init__.py`: CLI command definition and `asyncio` entry point.
    * `process.py`: Optimized logic for XML parsing, repository grouping, and file existence verification.
* **`tests/cli/cmd_check/test_process.py`**: Added comprehensive unit and integration tests using `pytest` and `unittest.mock`.
    * **`changelog.d/78.feature.rst`**: Added changelog fragment.

### DIFF
--- a/src/docbuild/cli/cmd_check/process.py
+++ b/src/docbuild/cli/cmd_check/process.py
@@ -93,4 +93,8 @@ async def process_check_files(
         results = await _verify_repository_files(url, branch, deli_list, repo_root)
         missing_files.extend(results)
 
+    # Final success message if no files were missing
+    if not missing_files:
+        log.info("All DC files are available in remote repositories.")
+
     return missing_files


src/docbuild/cli/cmd_check/process.py
```diff
@@ -0,0 +1,91 @@
...
+            branch = branch_node.text if branch_node is not None else "main"
+
+            repo = None
+            git_nodes = deli_node.xpath("ancestor::docset/builddocs/git")
+            if git_nodes:
+                repo = git_nodes[0].get("remote")
+
+            if repo and dcfile:
+                groups.setdefault((repo, branch), []).append(dcfile)
+        except Exception as e:
```
tomschr: Could we here more specific? Looking at the code, it's unlikely that we raise an exception here. We have guards and if-clauses already. Maybe remove that alltogether?

sushant-suse:
> Thank you very much! 👍
> 
> There is only one fix that wasn't applied. But exactly that one is very important, otherwise it doesn't work for me. 😆
> 
> Apart from that, I have one final question. If I don't add `-v`, `docbuild` doesn't output anything at all. It's probably okay and follows the Unix principle ("no news are good news"), but I'm not sure if we should have a more verbose view.
> 
> I don't think it's really needed. Perhaps we can leave it for now as it's functionally complete now. We can always improve later.
> 
> What do you think?

Thanks for the review @tomschr, I missed that `expanduser()`. Sorry about that.
- I have now applied the `.expanduser()` fix to both `config_dir` and `repo_root` to ensure tilde paths are handled correctly. That was definitely the 'missing link' for the environment setup.
- Regarding the verbosity, I agree with the Unix principle for the default view to keep it clean for CI/CD pipelines. However, to avoid total silence, I have ensured that if the check passes, the command exits cleanly. If you think it would be better, I can add a simple `log.info("All DC files verified successfully.")` at the very end of the process so users get at least one line of positive feedback without needing `-v`.

Something like this:
```python
    if not missing_files:
        log.info("All DC files are available in remote repositories.")
```

tomschr:
> Something like this:
> 
> ```python
>     if not missing_files:
>         log.info("All DC files are available in remote repositories.")
> ```

I was thinking more about using rich for output, outside of the normal logging. But for the time being, `log.info` is fine too. That's the easiest to implement it now.

Could you add that final two lines? Then we can merge it.

tomschr (approved):
Thanks a lot! Ready to be merged.

PR #174, opened 2026-02-15
Fix #171: Document ENV config TOML file

### LINKED ISSUE #171: Document the TOML ENV config file
### Affected Documentation Section

reference

### Suggested Change

The TOML example config is stored under `etc/docbuild/env.example.toml`. It contains several sections and key-value pairs that needs to be documented.

In the ideal case, we document a thorough description of the purpose, the datatypes, and examples.

### Additional Context

To document the keys, their types, and descriptions there a several methods:

* [Sphinx-Needs](https://sphinx-needs.readthedocs.io)
  This is one of the most powerful extensions for technical documentation. It has a `needs_from_toml` feature that allows you to pull configuration options directly from a TOML file and render them as "Requirements" or "Specifications" in your docs. This ensures your documentation always stays in sync with the actual config file.
 
* Custom TOML directive.
  If you just want to display a pretty-formatted TOML block with syntax highlighting, you don't need an extension. Sphinx supports it out of the box:

  ```rst
  .. code-block:: toml

   [docbuild]
   source_dir = "docs/"
   output_format = "html"
  ```

### Related

See also #170

tomschr:
* Generate the RST from TOML file through a Python script (`toml_parser.py`). It can be executed as a standalone script or is integrated into Sphinx.
* Enhance `conf.py` to read the TOML config(s) and create the generated file. The generated file is included in `reference/env-toml/index.rst`.
* The generated file has a different extension than `.rst` (actually `.rst.inc`). This was necessary otherwise you get a warning from Sphinx about "duplicate labels".
* Enrich the TOML file with comments similar to "docstrings": it uses the syntax `KEYNAME(TYPE): DESCRIPTION` as a "self-documenting" approach.
* Currently, the standard tomllib library doesn't preserve comments. To not rely on a different library, we need to read it line by line. :disappointed: 
* There is a small optimization in the `conf.py` file: the `.rst.inc` file is only generated when the TOML file has a newer date.

Additional changes that were needed:

* Introduce a new `github-action` group in `pyproject.toml`. This requires `tests` and `docs` as `conf.py` requires Sphinx now.
* Adjust the `workflows/ci.yml` to use the new `github-action` group.
* Update the `uv.lock` file to contain this new group too.

## Screenshot

Screenshot_20260215_113550

Meaning of the callouts in the image:

1. The title from the `reference/env-toml/index.rst`
2. The introductory paragraph, also from `reference/env-toml/index.rst`
3. The TOML section taken from the `etc/docbuild/env.example.toml`
4. The description of the section taken from a comment of `etc/docbuild/env.example.toml`
5. Also a comment before the key/value pair from `etc/docbuild/env.example.toml`

### DIFF
--- a/docs/source/toml_parser.py
+++ b/docs/source/toml_parser.py
@@ -54,7 +54,7 @@ logger = logging.getLogger(__name__)
 # This pattern defines a single TOML key part, which can be a bare key (unquoted),
 # a double-quoted key, or a single-quoted key. This is the building block for
 # more complex dotted keys.
-KEY_PART_PATTERN = r"""[A-Za-z0-9_-]+|"(?:\\.|[^"\\])*"|\'[^\']*\'"""
+KEY_PART_PATTERN = r"""[A-Za-z0-9_-]+|"(?:\\.|[^"\\])*"|'[^']*'"""
 
 # This verbose regex pattern matches a full TOML key (bare, quoted, or dotted).
 # A TOML key is composed of one or more "key parts" separated by dots.
@@ -73,8 +73,8 @@ DOTTED_KEY_PATTERN = fr"""
 
 COMMENT_METADATA_REGEX = re.compile(
     fr"""
-        ^\s*#\s*                                # Leading whitespace and comment marker
-        (?P<key>{DOTTED_KEY_PATTERN})          # The full TOML key (captured)
+        ^\s*[#]\s*                                # Leading whitespace and comment marker
+        (?P<key>{DOTTED_KEY_PATTERN})           # The full TOML key (captured)
         \((?P<type>[^)]+)\)                     # The type in parentheses (captured)
         :\s*                                    # Colon and optional whitespace
         (?P<description>.*)$                    # The rest of the line is the description (captured)


docs/source/toml_parser.py
```diff
@@ -0,0 +1,294 @@
...
+from typing import Any
+
+logger = logging.getLogger(__name__)
+
+# Regex patterns for TOML parsing
+COMMENT_METADATA_REGEX = re.compile(
+    r"^\s*#\s*(?P<key>[\w.-]+)\((?P<type>[^)]+)\):\s*(?P<description>.*)$"
+)
+TOML_KEY_VALUE_REGEX = re.compile(r"^\s*(?P<key>[\w.-]+)\s*=\s*.*$")
+TOML_SECTION_REGEX = re.compile(r"^\s*\[(?P<name>[\w.-]+)\]\s*$")
```
sushant-suse: Hi Toms, I think that the `TOML_SECTION_REGEX` and `TOML_KEY_VALUE_REGEX` are quite strict. If a user puts a comment on the same line as a key (for example, `port = 80 # The port`), these regexes might fail to match. Since this is specifically for a controlled "example" file, this is not a dealbreaker, but it is worth noting. WDYT?

docs/source/toml_parser.py
```diff
@@ -0,0 +1,294 @@
+"""TOML to reStructuredText (RST) Documentation Generator.
```
sushant-suse: The hybrid approach of using `tomllib` for values and manual parsing for comments is a very smart way to keep the documentation build dependency-free. The recursive lookup for nested sections (`paths.tmp`) is well-handled. One small suggestion for cross-platform reliability: Consider adding `encoding="utf-8"` to the `read_text()` and `write_text()` calls to ensure the parser behaves identically on Windows and Linux/macOS environments.

sushant-suse:
> * Generate the RST from TOML file through a Python script (`toml_parser.py`). It can be executed as a standalone script or is integrated into Sphinx.
> * Enhance `conf.py` to read the TOML config(s) and create the generated file. The generated file is included in `reference/env-toml/index.rst`.
> * The generated file has a different extension than `.rst` (actually `.rst.inc`). This was necessary otherwise you get a warning from Sphinx about "duplicate labels".
> * Enrich the TOML file with comments similar to "docstrings": it uses the syntax `KEYNAME(TYPE): DESCRIPTION` as a "self-documenting" approach.
> * Currently, the standard tomllib library doesn't preserve comments. To not rely on a different library, we need to read it line by line. 😞
> * There is a small optimization in the `conf.py` file: the `.rst.inc` file is only generated when the TOML file has a newer date.
> 
> Additional changes that were needed:
> 
> * Introduce a new `github-action` group in `pyproject.toml`. This requires `tests` and `docs` as `conf.py` requires Sphinx now.
> * Adjust the `workflows/ci.yml` to use the new `github-action` group.
> * Update the `uv.lock` file to contain this new group too.
> 
> ## Screenshot
> Screenshot_20260215_113550
> Meaning of the callouts in the image:
> 
> 1. The title from the `reference/env-toml/index.rst`
> 2. The introductory paragraph, also from `reference/env-toml/index.rst`
> 3. The TOML section taken from the `etc/docbuild/env.example.toml`
> 4. The description of the section taken from a comment of `etc/docbuild/env.example.toml`
> 5. Also a comment before the key/value pair from `etc/docbuild/env.example.toml`

Hi @tomschr thanks a lot for the awesome work.

The "Documentation as Code" approach is excellent. Syncing the documentation directly with `env.example.toml` ensures the reference guide never goes stale. Small feedback(s):
* **`toml_parser.py`**: Add `encoding="utf-8"` to `read_text()` and `write_text()` calls for cross-platform (Windows) reliability.
* **`conf.py`**: Fix the filename typo in the comment (`toml_doc_generator.py`  `toml_parser.py`).
* **`env.example.toml`**: The new `key(type): description` syntax is clean and well-applied across all sections.

Great engineering to keep the build dependency-free. Approved.

docs/source/toml_parser.py
tomschr: You're right. I've implemented that in commit 5def09d

Issue #183, opened 2026-02-18
Compare generated and manually JSON files


tomschr:
### Situation

Currently, our internal [configuration repo](https://gitlab.suse.de/susedoc/docserv-config/-/tree/main/json-portal-dsc) contains many JSON files that were crafted manually.

The goal is to have these JSON files generated by docbuild.

### Use Case

Writers don't need to care of all the details of having a valid and syntactically correct JSON file. In the past it was a challenge to provide changes not only to the Docserv XML config, but also reflect these changes manually in the JSON.

With the help of docbuild, this tedious task becomes obsolete.

### Possible Implementation

1. Clone the [GL://susedoc/docserv-config](https://gitlab.suse.de/susedoc/docserv-config/).
2. Pick one JSON file from the `json-portal-dsc` folder, for example `sles/16.0.json`.
3. Run `docbuild` to generate the metadata:

        docbuild --env-config=env.devel.toml metadata `sles/16.0/en-us''

   The result will be saved, for example, in `/var/cache/docbuild/doc-example-com/meta/sles/16.0.json`.
4. Compare the two files side-by-side. Either manually or by an appropriate tool (maybe https://jsondiff.com/?)

Differences can occur:

* A different order in `.documents`. That should be fine.
* A different key would be an error.
* A different value could be an error. That depends on the datatype.

We need to collect the differences to ensure they are the same. They don't have to be byte-identical, but have the same structure, keys etc.

Issue #126, opened 2026-01-07
Implement a strategy to select between native and containerized daps


tomschr:
#### Terms and Definitions
 
* **Native daps**
  * DAPS is usually via RPM (in very rare cases, you can install it from the GitHub repo).
  * All dependencies are covered by RPM.
  * The docbuild script runs daps directly (command needs to be accessible in `$PATH`).
 
* **Containerized daps**
  * Daps is installed inside a Docker image.
  * All dependencies are covered in the Docker image.
  * The docbuild script runs daps indirectly through `docker run`.
  

#### Docserv Background
The old Docserv code knew only one way: building documentation through a container that contains daps. Running daps natively wasn't available.

### Use Case

Each of the different ways have pros and cons. For a stable, reproducible build the Docker approach may be more sufficient.

For a testing purposes without the hassle of going through all the loops of updating the RPM packages and building the Docker image, a more direct approach is more feasible.

Combining the two makes the script more flexible.

### Possible Implementation

The classic approach is to implement the [strategy pattern](https://refactoring.guru/design-patterns/strategy).

<details>
   <summary>Classic Class-based Strategy pattern implementation</summary>

```python
from abc import ABC, abstractmethod
import subprocess

# 1. The Abstract Strategy (The Blueprint)
# Every specific method must follow this blueprint.
class DapsBuilderStrategy(ABC):
    """
    The interface that all versions of the daps runner must implement.
    """
    
    @abstractmethod
    def build(self, source_path: str, output_path: str) -> None:
        """
        Builds the documentation.
        
        :param source_path: Path to the documentation source (DC file).
        :param output_path: Where the built files should go.
        """
        pass

# 2. The Concrete Strategies (The Implementations)
class NativeDapsBuilder(DapsBuilderStrategy):
    def build(self, source_path: str, output_path: str) -> None:
        print(f"Build Mode: Native System")
        
        # Construct the native command
        # Example: daps -d DC-file build path
        command = ["daps", "-d", source_path, "--build", output_path]
        
        print(f"Executing: {' '.join(command)}")
        try:
            # subprocess.run is great for running shell commands
            subprocess.run(command, check=True)
            print("Native build successful.")
        except subprocess.CalledProcessError as e:
            print(f"Error running native daps: {e}")

class DockerDapsBuilder(DapsBuilderStrategy):
    def __init__(self, docker_image: str = "suse/doc-daps"):
        self.docker_image = docker_image

[...]

PR #184, opened 2026-02-18
feat: implement manifest parity with legacy JSON and fix nested collection

sushant-suse:
Fixes: #183 

### Key Changes:
- **Recursive Collection**: Fixed a bug where metadata files (DC-*) were not collected from nested product subdirectories.
- **Structural Parity**: Updated the Manifest Pydantic model with hyphenated aliases (e.g., `hide-productname`) to align with existing portal expectations.
- **Infrastructure**: Added support for containerized DAPS execution, enabling development on macOS environments.
- **Linguistic Drift Resolution**: Implemented automated "parity fixes" to ensure consistency in HTML entities and instructional paragraphs across all translated descriptions.

### Verification:

- Automated audit using DeepDiff confirms structural parity with legacy manifests.
- Passed all 660+ existing unit tests.
- Maintained 97.6% code coverage.
- Validated with `ruff` linting.

### DIFF
--- a/src/docbuild/cli/cmd_metadata/metaprocess.py
+++ b/src/docbuild/cli/cmd_metadata/metaprocess.py
@@ -292,8 +292,10 @@ async def process_doctype(
     # Complexity reduced by delegating task execution to the helper
     return await _run_metadata_tasks(tasks, deliverables, exitfirst)
 
-def _apply_parity_fixes(descriptions: list, categories: list) -> None:
+def apply_parity_fixes(descriptions: list, categories: list) -> None:
     """Apply wording and HTML parity fixes for legacy JSON consistency."""
+    # TODO: These strings are hard-coded for legacy parity but should be moved to
+    # Docserv config files to allow for proper translation and localization.
     legacy_tail = (
         "<p>The default view of this page is the ```Table of Contents``` sorting order. "
         "To search for a particular document, you can narrow down the results using the "
@@ -368,7 +370,7 @@ def store_productdocset_json(
         # 1. Capture and Clean Descriptions/Categories using helper
         descriptions = list(Description.from_xml_node(productnode))
         categories = list(Category.from_xml_node(productnode))
-        _apply_parity_fixes(descriptions, categories)
+        apply_parity_fixes(descriptions, categories)
 
         # 2. Initialize Manifest
         manifest = Manifest(


src/docbuild/cli/cmd_metadata/metaprocess.py
```diff
@@ -111,99 +155,52 @@ async def process_deliverable(
     """
     log.info("> Processing deliverable: %s", deliverable.full_id)
 
-    meta_cache_dir = Path(meta_cache_dir)
+    if not context.envconfig:
+        log.error("Environment configuration is missing.")
+        return False
```
tomschr: Is this really needed? Can't we assume it's there?

If we call it by `docbuild metadata`, the `.envconfig` is a valid `EnvConfig` object. If not, there is something wrong with the config.

src/docbuild/cli/cmd_metadata/metaprocess.py
tomschr: When you pass the context, the other arguments becomes obsolete as they can derived from the former, right?

That would mean, we reduce the arguments to this:

```suggestion
dapstmpl: str,
```

src/docbuild/cli/cmd_metadata/metaprocess.py
```diff
@@ -111,99 +155,52 @@ async def process_deliverable(
     """
     log.info("> Processing deliverable: %s", deliverable.full_id)
 
-    meta_cache_dir = Path(meta_cache_dir)
+    if not context.envconfig:
```
tomschr: See also docstring:

```suggestion
    env = context.envconfig
    repo_dir = env.paths.repo_dir
    tmp_repo_dir = env.paths.tmp_repo_dir
    base_cache_dir = env.paths.base_cache_dir
    meta_cache_dir = env.paths.meta_cache_dir
```

src/docbuild/cli/cmd_metadata/metaprocess.py
```diff
@@ -54,33 +53,77 @@ def collect_files_flat(
...
+            f for f in all_files
+            if dt.product.value.lower() in [p.lower() for p in f.parts]
+            and docset.lower() in [p.lower() for p in f.parts]
         ]
 
-        # Only yield if we found something
         if files:
             yield dt, docset, files
 
+def _get_daps_command(
```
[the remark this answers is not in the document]
sushant-suse: You're right, the containerized execution logic does overlap with #126. I originally included it to facilitate development and testing on macOS where a native DAPS installation is difficult.

To keep this PR focused strictly on the the current metadata parity requirements, I can revert the `_get_daps_command` to use only the native template. We can then implement the full Strategy Pattern as part of a dedicated PR for #126. Does that work for you?

src/docbuild/cli/cmd_metadata/metaprocess.py
sushant-suse: You're right. Since `docbuild metadata` ensures that `EnvConfig` is initialized and validated before calling the processing logic, this check is redundant. I've removed it to keep the function clean.

src/docbuild/cli/cmd_metadata/metaprocess.py
sushant-suse: That's a great point. Since we are already passing the context object, extracting these paths within the function instead of passing them as individual arguments makes the signature much cleaner and less redundant. I have updated the function to derive `repo_dir`, `tmp_repo_dir`, `base_cache_dir`, and `meta_cache_dir` directly from `context.envconfig.paths`.

src/docbuild/cli/cmd_metadata/metaprocess.py
```diff
@@ -320,88 +318,107 @@ async def process_doctype(
...
+                loaded_doc_data = json.load(fh)
+
+            if not loaded_doc_data:
+                log.warning("Empty metadata file %s", f)
+                continue
+
+            doc_model = Document.model_validate(loaded_doc_data)
+            manifest.documents.append(doc_model)
+
+        except Exception as e:
```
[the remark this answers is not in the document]
sushant-suse: I agree. Catching a broad `Exception` is too vague. I have updated the logic to specifically catch `json.JSONDecodeError` for malformed files, `ValidationError` for schema mismatches, and `OSError` for filesystem issues. This allows us to provide more targeted error messages while still skipping only the problematic file.

src/docbuild/cli/cmd_metadata/metaprocess.py
```diff
@@ -95,118 +117,70 @@ async def process_deliverable(
...
-            # stdout.print(f'  command: {cmd}')
-            daps_process = await asyncio.create_subprocess_exec(
-                *cmd,
-                stdin=asyncio.subprocess.DEVNULL,
-                stdout=asyncio.subprocess.PIPE,
-                stderr=asyncio.subprocess.PIPE,
+
+            daps_proc = await asyncio.create_subprocess_exec(
+                *cmd, stdin=asyncio.subprocess.DEVNULL,
+                stdout=asyncio.subprocess.PIPE, stderr=asyncio.subprocess.PIPE,
```
tomschr: Technically correct, but I think it's visually more readable when each part is separated like this:

```suggestion
            daps_proc = await asyncio.create_subprocess_exec(
                *cmd,
                stdin=asyncio.subprocess.DEVNULL,
                stdout=asyncio.subprocess.PIPE,
                stderr=asyncio.subprocess.PIPE,
```

src/docbuild/cli/cmd_metadata/metaprocess.py
```diff
@@ -54,39 +54,61 @@ def collect_files_flat(
...
+    """Construct the DAPS command for native execution."""
+    raw_daps_cmd = dapstmpl.format(
+        builddir=str(worktree_dir),
+        dcfile=str(dcfile_path),
+        output=str(outputjson),
+    )
+    return shlex.split(raw_daps_cmd)
+
+
+def _update_metadata_json(outputjson: Path, deliverable: Deliverable) -> None:
```
tomschr: Remove the underscore:

```suggestion
def update_metadata_json(outputjson: Path, deliverable: Deliverable) -> None:
```

src/docbuild/cli/cmd_metadata/metaprocess.py
```diff
@@ -244,164 +263,148 @@ async def process_doctype(
...
         product = doctype.product.value
-        stdout.print(f" ❱ Processed group: {doctype} / {docset}")
-        # The XPath logic is encapsulated within the Doctype model
+
+        # Version Formatting (Safety Fix)
+        # Only remove .0 if the docset appears to be a version number
+        # and not a purely alphanumeric identifier (Fixing dangerous remsuffix)
+        version_str = str(docset)
+        if version_str.replace('.', '', 1).isdigit() and version_str.endswith(".0"):
+            version_str = version_str.removesuffix(".0")
```
tomschr: I'm sorry, this is not the correct handling. Better don't touch the version string at all.

src/docbuild/cli/cmd_metadata/metaprocess.py
```diff
@@ -244,164 +263,148 @@ async def process_doctype(
...
-
-        #
-        jsonfile.write_text(manifest.model_dump_json(indent=2, by_alias=True))
-        log.info(
-            "Wrote merged metadata JSON for %s/%s => %s", product, docset, jsonfile
-        )
+        jsonfile = jsondir / f"{docset}.json"
+
+        # Exporting with aliases to maintain parity with legacy JSON keys
+        json_data = manifest.model_dump(by_alias=True, exclude_none=True)
```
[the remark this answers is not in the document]
sushant-suse: Yes, removed it.

src/docbuild/cli/cmd_metadata/metaprocess.py
sushant-suse: > Understood. I have removed the version string manipulation. The code now uses the raw `docset` value directly as the `version_str` without any suffix removal or digit checks.

PR #194, opened 2026-02-24
Fix #191: Add producer-consumer model

### LINKED ISSUE #191: Abstract producer/consumer model
### Situation

We want to apply the producer/consumer model on different places. It would be helpful to keep it DRY.

### Use Case

Avoid boilerplate code

### Possible Implementation

In most case, it's the same boilerplate code:

```python
# src/docbuild/utils/concurrency.py
import asyncio
import logging
from typing import TypeVar, Callable, Awaitable, Iterable

T = TypeVar("T")  # Input type
R = TypeVar("R")  # Result type

log = logging.getLogger(__name__)

async def parallel_process(
    items: Iterable[T],
    worker_fn: Callable[[T], Awaitable[R]],
    limit: int,
    return_exceptions: bool = False,
) -> list[R | Exception]:
    """
    Process a list of items in parallel using a fixed number of workers.

    :param items: An iterable of items to process.
    :param worker_fn: An async function that processes a single item.
    :param limit: The maximum number of concurrent workers.
    :param return_exceptions: If True, exceptions are returned as results 
                              instead of raised.
    :return: A list of results (unordered unless you track indices).
    """
    queue: asyncio.Queue[T] = asyncio.Queue()
    results: list[R | Exception] = []
    
    # 1. Populate Queue
    for item in items:
        queue.put_nowait(item)
    
    # 2. Define Worker
    async def worker() -> None:
        while True:
            try:
                item = await queue.get()
            except asyncio.CancelledError:
                return

            try:
                res = await worker_fn(item)
                results.append(res)
            except Exception as e:
                if return_exceptions:
                    results.append(e)
                else:
                    log.error("Worker failed: %s", e)
                    # Optional: Cancel all other workers here if you want fail-fast
            finally:
                queue.task_done()

    # 3. Start Workers
    workers = [asyncio.create_task(worker()) for _ in range(limit)]

    # 4. Wait & Cleanup
    await queue.join()
    for w in workers:
        w.cancel()
    await asyncio.gather(*workers, return_exceptions=True)

    return results
```

tomschr:
This PR plays around with the producer-consumer model requested in issue #191. It maybe be integrated or it goes nowhere.

To test it, use these commands:

```shell
source devel/activate-aliases.sh
upython -m docbuild.utils.concurrency
```

Questions:

* ~The current implementation uses `TaskGroup`s, which is considered as the go-to way. Semaphores seems not always that reliable or appropriate for this use case.~ 
   => We use `TaskGroup`s
* ~What do we want to "return" inside the `consumer()` loop?
  The consumer calls the worker function. Depending on if it's successful or not, we add either a `Success` object or a `Failure` object.
  I'm not sure if this is a bit too much and can be simplified.~
  => We return the respective object and if something fails, a `TaskFailedError` exception.
* I'm not sure about the naming: `process_unordered`. Possible better names:
  * `execute_concurrently`
  * `run_in_parallel`

### DIFF
--- a/src/docbuild/utils/concurrency.py
+++ b/src/docbuild/utils/concurrency.py
@@ -3,6 +3,7 @@
 import asyncio
 from collections.abc import Awaitable, Callable, Iterable
 import logging
+from typing import Concatenate, ParamSpec
 
 log = logging.getLogger(__name__)
 
@@ -20,9 +21,12 @@ class TaskFailedError[T](Exception):
         self.original_exception = original_exception
 
 
+async def process_unordered[T, R, **P](
     items: Iterable[T],
-    worker_fn: Callable[[T], Awaitable[R]],
+    worker_fn: Callable[Concatenate[T, P], Awaitable[R]],
     limit: int,
+    *worker_args: P.args,
+    **worker_kwargs: P.kwargs,
 ) -> list[R | TaskFailedError[T]]:
     """Process items concurrently with a worker limit.
 
@@ -32,7 +36,10 @@ class TaskFailedError[T](Exception):
 
     :param items: Iterable of items to process.
     :param worker_fn: Async function processing a single item.
+        Result signature: `worker_fn(item, *worker_args, **worker_kwargs)`.
     :param limit: Max concurrent workers.
+    :param worker_args: Additional positional arguments passed to `worker_fn`.
+    :param worker_kwargs: Additional keyword arguments passed to `worker_fn`.
     """
     # Limit queue size to prevent memory explosion if producer is faster than consumers
     queue: asyncio.Queue[T | None] = asyncio.Queue(maxsize=limit * 2)
@@ -50,7 +57,7 @@ class TaskFailedError[T](Exception):
                 break
 
             try:
-                result_val = await worker_fn(item)
+                result_val = await worker_fn(item, *worker_args, **worker_kwargs)
                 results.append(result_val)
 
             except Exception as e:
--- a/src/docbuild/utils/concurrency.py
+++ b/src/docbuild/utils/concurrency.py
@@ -1,8 +1,7 @@
-
-# src/docbuild/utils/concurrency.py
-from multiprocessing.sharedctypes import Value
+from abc import ABC, abstractmethod
 import asyncio
 from collections.abc import Awaitable, Callable, Iterable
+from dataclasses import dataclass
 import logging
 from typing import TypeVar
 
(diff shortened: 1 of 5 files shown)


sushant-suse:
> This PR plays around with the producer-consumer model requested in issue #191. It maybe be integrated or it goes nowhere.
> 
> To test it, use these commands:
> 
> ```shell
> source devel/activate-aliases.sh
> upython -m docbuild.utils.concurrency
> ```
> 
> Questions:
> 
> * ~The current implementation uses `TaskGroup`s, which is considered as the go-to way. Semaphores seems not always that reliable or appropriate for this use case.~
>   => We use `TaskGroup`s
> * ~What do we want to "return" inside the `consumer()` loop?
>   The consumer calls the worker function. Depending on if it's successful or not, we add either a `Success` object or a `Failure` object.
>   I'm not sure if this is a bit too much and can be simplified.~
>   => We return the respective object and if something fails, a `TaskFailedError` exception.
> * I'm not sure about the naming: `process_unordered`. Possible better names:
>   
>   * `execute_concurrently`
>   * `run_in_parallel`

Thanks for the awesome work Toms. It looks great. I have also added my suggestions.

src/docbuild/utils/concurrency.py
```diff
@@ -0,0 +1,180 @@
...
+                # Wrap the exception in TaskFailedError
+                results.append(TaskFailedError(item, e))
+
+            finally:
+                queue.task_done()
+
+    async with asyncio.TaskGroup() as tg:
+        # Start consumers
+        for _ in range(limit):
+            tg.create_task(consumer())
```
sushant-suse: Hi Toms, currently the `tg.create_task(consumer())` and `await producer()` are both inside the `TaskGroup` block. If the `producer()` fails (for example, an error while iterating over items), the `TaskGroup` will immediately try to shut down. However, the consumers are stuck waiting on `queue.get()`. Should we use a `try...finally` or ensure the "Sentinel" (the `None` value) is sent even if the producer fails?

src/docbuild/utils/concurrency.py
```diff
@@ -0,0 +1,180 @@
...
+        await producer()
+
+        # Signal shutdown
+        for _ in range(limit):
+            await queue.put(None)
+
+        # Wait for all items to be processed
+        await queue.join()
+
+    return results
```
sushant-suse: This is a great implementation of the Producer-Consumer pattern. It provides the backpressure and resource management we need. I have a few suggestions to make it a bit more robust for the core library:

1. **Avoid `None` as a sentinel**: If an item in the input iterable is actually `None`, the consumer will stop prematurely. Using a private `_DONE = object()` is safer.
2. **Naming**: `map_concurrently` might be more intuitive for users familiar with Python's `map()` or `multiprocessing.Pool.map()`.
3. **Refined Consumer**: I'd suggest explicitly handling `asyncio.CancelledError` in the worker loop so it doesn't get wrapped in a `TaskFailedError` during a normal shutdown.

For the `async def producer()`, I moved that loop directly into the `TaskGroup`. This is cleaner because it removes one extra nested function.

```suggestion
# Create a unique sentinel to avoid conflicts if an item is naturally None
_DONE = object()

async def map_concurrently[T, R, **P](
    items: Iterable[T],
    worker_fn: Callable[Concatenate[T, P], Awaitable[R]],
    limit: int,
    *worker_args: P.args,
    **worker_kwargs: P.kwargs,
) -> list[R | TaskFailedError[T]]:
    """Map a function over items concurrently with a fixed worker limit.

    Provides backpressure via a bounded queue and aggregates results
    as they complete. 
    """
    # maxsize=limit ensures we don't buffer too much in memory
    queue: asyncio.Queue[T | object] = asyncio.Queue(maxsize=limit)
    results: list[R | TaskFailedError[T]] = []

    async def consumer() -> None:
        while True:
            item = await queue.get()
            if item is _DONE:
                queue.task_done()
                break
            
            try:
                # Execute the worker
                res = await worker_fn(item, *worker_args, **worker_kwargs)
                results.append(res)
            except asyncio.CancelledError:
                raise # Propagate cancellation for TaskGroup
            except Exception as e:
                log.error("Task failed for item %s: %s", item, e)
                results.append(TaskFailedError(item, e))
            finally:
                queue.task_done()

    try:
        async with asyncio.TaskGroup() as tg:
            # 1. Spawn workers
            workers = [tg.create_task(consumer()) for _ in range(limit)]

            # 2. Feed the queue
            for item in items:
                await queue.put(item)

            # 3. Shutdown signaling
            for _ in range(limit):
                await queue.put(_DONE)
    except Exception as e:
        log.error("Producer-Consumer pipeline encountered an error: %s", e)
        raise

    return results
```

What do you think about using a private sentinel and the `map_` naming convention?

sushant-suse (changes requested):

src/docbuild/utils/concurrency.py
```diff
@@ -0,0 +1,180 @@
...
+
+import asyncio
+from collections.abc import Awaitable, Callable, Iterable
+import logging
+from typing import Concatenate
+
+log = logging.getLogger(__name__)
+
+
+class TaskFailedError[T](Exception):
```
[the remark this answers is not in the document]
tomschr: But that's the class, not the function. Inside `process_unordered`/`run_parallel` we have a `ParamSpec`.

sushant-suse (changes requested):

tomschr (commented):
Ready to be squash-merged! 👍 🎉

PR #221, opened 2026-04-21
fix(cli): handle TOML syntax errors gracefully

sushant-suse:
#### Core Logic & CLI
* **`src/docbuild/cli/cmd_cli.py`**:
    * Wrapped the `handle_config` calls for both **Application** and **Environment** configurations in `try...except` blocks to catch `tomllib.TOMLDecodeError`.
    * Updated `handle_validation_error` (and its internal logic) to distinguish between TOML syntax errors and Pydantic validation errors.
    * Ensured the specific filename causing the syntax error is passed to the error formatter to provide better context.
    * Initialized configuration dictionaries to prevent Pylance "possibly unbound" warnings.

#### Utilities
* **`src/docbuild/utils/errors.py`**:
    * Added a new `format_toml_error` function. This uses the `rich` library to print a clean, red-styled error panel.
    * The panel includes the specific error message from the parser (including line/column) and a helpful hint about TOML boolean casing.

#### Tests
* **`tests/utils/test_errors.py`**: Added a test to verify that the `format_toml_error` utility correctly captures and formats a syntax error.
* **`tests/cli/test_cmd_cli.py`**: Added an end-to-end test that mocks a `TOMLDecodeError` during the CLI initialization to verify the program exits with code `1` and displays the correct error message without a traceback.

### DIFF
--- a/src/docbuild/cli/cmd_cli.py
+++ b/src/docbuild/cli/cmd_cli.py
@@ -91,11 +91,9 @@ def load_app_config(
 ) -> None:
     """Load and validate Application configuration.
 
-    Args:
-        ctx: The Click context object.
-        app_config: The path to the application config file provided via CLI.
-        max_workers: The max_workers value from CLI options.
-
+    :param ctx: The Click context object. The result will be added to ``ctx.obj.appconfig``.
+    :param app_config: The path to the application config file provided via CLI.
+    :param max_workers: The max_workers value from CLI options.
     """
     context = ctx.obj
     result = handle_config(
@@ -118,10 +116,8 @@ def load_app_config(
 def load_env_config(ctx: click.Context, env_config: Path) -> None:
     """Load and validate Environment configuration.
 
-    Args:
-        ctx: The Click context object.
-        env_config: The path to the environment config file provided via CLI.
-
+    :param ctx: The Click context object. The result will be added to ``ctx.obj.envconfig``.
+    :param env_config: The path to the environment config file provided via CLI.
     """
     context = ctx.obj
     result = handle_config(
--- a/src/docbuild/cli/cmd_cli.py
+++ b/src/docbuild/cli/cmd_cli.py
@@ -84,10 +84,9 @@ def handle_validation_error(
     ctx.exit(1)
 
 
-def _load_app_config(
+def load_app_config(
     ctx: click.Context,
     app_config: Path,
-    verbose: int,
     max_workers: str | None
 ) -> None:
     """Load and validate Application configuration.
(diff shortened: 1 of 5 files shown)


src/docbuild/cli/cmd_cli.py
```diff
@@ -68,21 +69,98 @@ def handle_validation_error(
...
     else:
-        log.error("%s configuration failed validation:", config_label)
+        config_label = "Application" if model_class == AppConfig else "Environment"
+        log.error("%s configuration failed:", config_label)
         log.error("Error in config file(s): %s", config_files)
         log.error(e)
     ctx.exit(1)
 
 
+def _load_app_config(
```
tomschr: I would use a name without the underscore as prefix:


```suggestion
def load_app_config(
```

Apart from this minor detail, the function is quite similar to `load_env_config`. It kind of violates the DRY principle and it concerns me.

Maybe you could move the `try...except` block one level up, to the caller?

The idea is to pretend everything would be okay:

```python
def _load_app_config(
    ctx: click.Context,
    app_config: Path,
    verbose: int,
    max_workers: str | None
) -> None:
    result = handle_config(
        app_config,
        CONFIG_PATHS,
        APP_CONFIG_BASENAMES + PROJECT_LEVEL_APP_CONFIG_FILENAMES,
        None,
        DEFAULT_APP_CONFIG,
    )
    context.appconfigfiles, raw_appconfig, context.appconfig_from_defaults = cast(
        tuple[tuple[Path, ...] | None, dict[str, Any], bool], result
    )
    if max_workers is not None:
        raw_appconfig["max_workers"] = max_workers
```

The error handling would be the `cli` function:

```python
# ... all @click 
def cli(...):
    try:
       # 1. Load Application Config
       configtype = AppConfig
       files = (app_config,) if app_config else None
       load_app_config(ctx, app_config, verbose, max_workers)
       # 2. Setup logging ...

       # 3. Load Environment Config
       configtypes = EnvConfig
       files = (env_config,) if env_config else None
       load_env_config(ctx, env_config, verbose)

       # 4. Setup Concurrency Lock...

   except (ValueError, ValidationError, tomllib.TOMLDecodeError) as e:
       handle_validation_error(e, configtype, files, verbose, ctx)
```

I'm not completely sure if this is the way to go. But it would leave the `load_{app,env}_config` functions with its primary goal to prepare everything.

All that said, I think there is another idea... maybe we don't need these separate functions at all? Could we use Pydantic for all the validation details (like this `max_workers`)? The two functions looks like an anti pattern to me.

src/docbuild/cli/cmd_cli.py
```diff
@@ -68,21 +69,98 @@ def handle_validation_error(
...
+    if max_workers is not None:
+        raw_appconfig["max_workers"] = max_workers
+
+    try:
+        context.appconfig = AppConfig.from_dict(raw_appconfig)
+    except (ValueError, ValidationError) as e:
+        handle_validation_error(e, AppConfig, context.appconfigfiles, verbose, ctx)
+
+
+def _load_env_config(ctx: click.Context, env_config: Path, verbose: int) -> None:
```
tomschr: Same here:


```suggestion
def load_env_config(ctx: click.Context, env_config: Path, verbose: int) -> None:
```

src/docbuild/utils/errors.py
```diff
@@ -112,3 +114,30 @@ def format_pydantic_error(
             f"[dim]... and {error_count - max_display} more errors. "
             "Use '-vv' to see all errors.[/dim]\n"
         )
+
+def format_toml_error(
```
tomschr, line 118: This is great! 👍

tomschr (changes requested):
Thanks @sushant-suse for the work. 👍 It took a bit longer for me.

Overall it looks good. 👍 Apart from minor details, I have some concerns regarding an anti pattern (see below). Not sure if we want to fix it here in this PR as it deviates from the original issue.

src/docbuild/cli/cmd_cli.py
```diff
@@ -68,21 +69,74 @@ def handle_validation_error(
...
+    ctx: click.Context,
+    app_config: Path,
+    max_workers: str | None
+) -> None:
+    """Load and validate Application configuration.
+
+    Args:
+        ctx: The Click context object.
+        app_config: The path to the application config file provided via CLI.
+        max_workers: The max_workers value from CLI options.
```
tomschr: These entries should be in Spinx RST style, otherwise our Autodoc could produce the wrong results:

```suggestion
    :param ctx: The Click context object. The result will be added to ``ctx.obj.appconfig``.
    :param app_config: The path to the application config file provided via CLI.
    :param max_workers: The max_workers value from CLI options.
```

src/docbuild/cli/cmd_cli.py
```diff
@@ -68,21 +69,74 @@ def handle_validation_error(
...
+
+    context.appconfig = AppConfig.from_dict(raw_appconfig)
+
+
+def load_env_config(ctx: click.Context, env_config: Path) -> None:
+    """Load and validate Environment configuration.
+
+    Args:
+        ctx: The Click context object.
+        env_config: The path to the environment config file provided via CLI.
```
tomschr: Same here:


```suggestion
    :param ctx: The Click context object. The result will be added to ``ctx.obj.envconfig``.
    :param env_config: The path to the environment config file provided via CLI.
```

tomschr (changes requested):
Just a tiny addition, otherwise it's great. 👍 Thank you! I think, it's ready to be squash-merged to main.

src/docbuild/cli/cmd_cli.py
sushant-suse: Oops. Forgot and used normal approach.

sushant-suse:
> Just a tiny addition, otherwise it's great. 👍 Thank you! I think, it's ready to be squash-merged to main.

Also forgot to add fragment file. Added that as well.

tomschr (approved):
Thanks! 👍

Issue #112, opened 2025-12-03
Rework `config` subcommand


tomschr:
### Situation

The `config` subcommand contains two other subcommands:

```
$ docbuild config -h
Usage: docbuild config [OPTIONS] COMMAND [ARGS]...

  CLI interface to shows config files how docbuild sees it.

Options:
  -h, --help  Show this message and exit.

Commands:
  app  CLI interface to show the configuration of the application files.
  env  CLI interface to showsthe configuration of the environment files.
```

Maybe we shouldn't distinguish between env and app subcommand? Or maybe use a specific option and not a subcommand?

### Use Case

In m humble opinion, the `config` subcommand should be used for:

* validating the config and showing validation errors
* showing the config. Especially when the env config contains placeholders the resolved paths are a great tool to debug if everything was correct.
* Something else?

### Possible Implementation

Maybe we should move from the distinction between app/env to a more use case oriented subcommand?

For example, we could implement:

* `config validate`
  We would need to answer _where_ do we want to pass the config file(s). At the moment, for the ENV config we have the options `--env-config`/`--app-config`. But passing that, triggers the validation already. If you have a wrong config and want to show the help, the help output won't be shown as validation errors are.
 
* `config show|list`
  This prints the complete configuration as JSON. All JSON should go to stdout to make it possible to pass it to other tools (e.g. `jq`).
  Maybe `config list` is nice as there is already a `git config list` command.
  Additionally, we could think about "flattening" the config like `git config list` does. That would require an option `--flat` or similar.

PR #230, opened 2026-04-23
refactor #112: reorganize config CLI and fix help-mode crash

sushant-suse:
Related Issue #112 

### Key Changes

1. Lazy-Loading & Help Logic (`src/docbuild/cli/cmd_cli.py`)
Implemented a "Help Shield": The root `cli` function now detects if a help flag (`--help` or `-h`) is present in `sys.argv` before attempting to load any configuration files.

2. Task-Oriented Config Subcommands
Merged Resources: Replaced the separate config application and config environment commands with unified config list and config validate commands. Added config list: Supports `--app`, `--env`, ~and `--portal`~ flags with an optional `--flat` output mode for "git-style" dotted configuration views. Added config validate: Centralizes validation logic for both application and environment files in one place.

3. Test Suite Stabilization
Metadata Logging: Updated `tests/cli/cmd_metadata/test_cmd_metadata.py` to use `unittest.mock.patch` on the module-level logger. This ensures tests passing by bypassing caplog capture issues in async environments.

Help Verification: Added `tests/cli/test_help.py` to specifically verify that `load_app_config` is never called when the help flag is passed.


src/docbuild/cli/cmd_config/list.py
```diff
@@ -0,0 +1,62 @@
...
+    items: list[tuple[str, Any]] = []
+    for k, v in d.items():
+        new_key = f"{prefix}.{k}" if prefix else k
+        if isinstance(v, dict):
+            items.extend(flatten_dict(v, new_key).items())
+        else:
+            items.append((new_key, v))
+    return dict(items)
+
+def _print_section(title: str, data: dict[str, Any], prefix: str, flat: bool, color: str) -> None:
```
tomschr: I would remove the underscore here as it's not really a very secret function. I would also introduce an additional line break as it's visually more readable (I think it should pass the ruff check).


```suggestion

def print_section(title: str, data: dict[str, Any], prefix: str, flat: bool, color: str) -> None:
```

src/docbuild/cli/cmd_config/list.py
```diff
@@ -0,0 +1,62 @@
...
+    """Print the Application and Environment configuration sections."""
+    if flat:
+        for k, v in flatten_dict(data, prefix).items():
+            console.print(f"[bold {color}]{k}[/bold {color}] = [green]{v}[/green]")
+    else:
+        console.print(f"\n# {title}", style="blue")
+        print_json(data=data)
+
+def _print_portal(doctypes: list[Any], flat: bool) -> None:
```
tomschr: Same here:


```suggestion

def print_portal(doctypes: list[Any], flat: bool) -> None:
```

src/docbuild/cli/cmd_config/list.py
```diff
@@ -0,0 +1,62 @@
...
+    """Print the Portal and Doctype metadata section."""
```
tomschr: Could you document the arguments in the docstring?

src/docbuild/cli/cmd_config/list.py
```diff
@@ -0,0 +1,62 @@
...
+    show_all = not (app or env or portal)
+
+    if (app or show_all) and context.appconfig:
+        _print_section("Application Configuration", context.appconfig.model_dump(), "app", flat, "cyan")
+
+    if (env or show_all) and context.envconfig:
+        _print_section("Environment Configuration", context.envconfig.model_dump(), "env", flat, "yellow")
+
+    if (portal or show_all) and context.doctypes:
+        _print_portal(context.doctypes, flat)
```
tomschr: Adjust the function names accordingly:

```suggestion
        print_section("Application Configuration", context.appconfig.model_dump(), "app", flat, "cyan")

    if (env or show_all) and context.envconfig:
        print_section("Environment Configuration", context.envconfig.model_dump(), "env", flat, "yellow")

    if (portal or show_all) and context.doctypes:
        print_portal(context.doctypes, flat)
```

src/docbuild/cli/cmd_config/validate.py
```diff
@@ -0,0 +1,57 @@
...
+                console.print(f"   [dim]- {f}[/dim]")
+        elif context.envconfig_from_defaults:
+            console.print("   [dim]- Using internal defaults[/dim]")
+
+    # 3. Portal/Doctype Status
+    if context.doctypes:
+        console.print(f"\n✅ [bold]Portals/Doctypes:[/bold] {len(context.doctypes)} discovered")
+        for doctype in context.doctypes:
+            name = getattr(doctype, "name", "Unknown")
+            console.print(f"   [dim]- {name}[/dim]")
```
tomschr: That's not validation. Validation is much more.😉

After looking at your code, I remembered that I had done something similar for the Docserv/Portal validation. I introduced a `docbuild validate` subcommand back then.
See `src/docbuild/cli/cmd_validate/process.py`.

Sorry that I didn't tell you first. I guess we should remove the `docbuild validate` subcommand. Perhaps you should also move these functions and move it here.

src/docbuild/cli/cmd_config/validate.py
tomschr, line 1: I feel like this needs to be broken down even further. Why not create a package like this:

```
validate
├── __init__.py
├── app.py
├── env.py
└── portal.py
```

The `__init__.py` would be the "orchestrator" and delegate it to the respective configs. Especially the `portal.py` will become quite big. This is not driven by Pydantic and needs a specific validation logic (using RELAX NG and some Python logic).

src/docbuild/cli/cmd_cli.py
```diff
@@ -219,7 +219,13 @@ def cli(
     context = ctx.obj
     context.verbose, context.dry_run, context.debug = verbose, dry_run, debug
 
-    # State tracking for centralized error handling
+    # --- LAZY FIX ---
+    # If the user is just asking for help, STOP HERE.
+    # Click will handle the help display for the subcommands automatically.
+    if ctx.resilient_parsing or "--help" in sys.argv or "-h" in sys.argv:
```
tomschr: Is it really needed to explicitly check for `--help`/`-h`? Shouldn't that be the task of Click? This looks like a code smell to me. 😉

src/docbuild/cli/cmd_config/list.py
```diff
@@ -0,0 +1,62 @@
...
+def flatten_dict(d: dict[str, Any], prefix: str = "") -> dict[str, Any]:
+    """Flatten a nested dictionary into dotted keys (e.g., app.logging.level)."""
+    items: list[tuple[str, Any]] = []
+    for k, v in d.items():
+        new_key = f"{prefix}.{k}" if prefix else k
+        if isinstance(v, dict):
+            items.extend(flatten_dict(v, new_key).items())
+        else:
+            items.append((new_key, v))
+    return dict(items)
```
tomschr: This function looks quite general. Maybe move this function somewhere in `src/docbuild/utils/`? Use `flatten.py`?

tomschr: Is converting the list to a dict really necessary? It should be left to the caller.
Additionally, maybe turn the function into a generator. This won't hold the entire list in memory, making it more efficient.

```python
def flatten_dict(d: dict[str, Any], prefix: str = "") -> Generator[tuple[str, Any], None, None]:
    """
    A generator that yields flattened key-value pairs from a nested dictionary.
    
    :param prefix: The accumulated path of keys from higher levels of 
       nesting. This is used to build the dotted string (e.g., "parent.child").
       Defaults to an empty string for the root level.
    :yields: A tuple containing the (dotted_key, value) pair.
    """
    for k, v in d.items():
        # Construct the new key based on the current nesting level
        new_key = f"{prefix}.{k}" if prefix else k
        
        if isinstance(v, dict):
            # yield from recursively calls the generator and yields its results
            yield from flatten_dict(v, new_key)
        else:
            # Yield the final flattened key and its non-dict value
            yield new_key, v
```

src/docbuild/cli/cmd_config/list.py
```diff
@@ -0,0 +1,62 @@
...
+def list_config(ctx: click.Context, app: bool, env: bool, portal: bool, flat: bool) -> None:
+    """List the configuration as JSON or flat text."""
+    context = ctx.obj
+    show_all = not (app or env or portal)
+
+    if (app or show_all) and context.appconfig:
+        _print_section("Application Configuration", context.appconfig.model_dump(), "app", flat, "cyan")
+
+    if (env or show_all) and context.envconfig:
+        _print_section("Environment Configuration", context.envconfig.model_dump(), "env", flat, "yellow")
```
tomschr: I'm not sure why, but when I use my `env.devel.toml` file, I get a traceback:

<details>
  <summary>Traceback of docbuild config list --env</summary>

```
$ docbuild --env-config=env.devel.toml config list --env

# Environment Configuration
  File "/home/toms/repos/GH/opensuse/docbuild.git/main/.venv/bin/docbuild", line 10, in <module>
    sys.exit(cli())
             ~~~^^
  File "/home/toms/repos/GH/opensuse/docbuild.git/main/.venv/lib/python3.14/site-packages/click/core.py", line 1514, in __call__
    return self.main(*args, **kwargs)
           ~~~~~~~~~^^^^^^^^^^^^^^^^^
  File "/home/toms/repos/GH/opensuse/docbuild.git/main/.venv/lib/python3.14/site-packages/click/core.py", line 1435, in main
    rv = self.invoke(ctx)
  File "/home/toms/repos/GH/opensuse/docbuild.git/main/.venv/lib/python3.14/site-packages/click/core.py", line 1902, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^
  File "/home/toms/repos/GH/opensuse/docbuild.git/main/.venv/lib/python3.14/site-packages/click/core.py", line 1902, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^
  File "/home/toms/repos/GH/opensuse/docbuild.git/main/.venv/lib/python3.14/site-packages/click/core.py", line 1298, in invoke
    return ctx.invoke(self.callback, **ctx.params)
           ~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/toms/repos/GH/opensuse/docbuild.git/main/.venv/lib/python3.14/site-packages/click/core.py", line 853, in invoke
    return callback(*args, **kwargs)
  File "/home/toms/repos/GH/opensuse/docbuild.git/main/.venv/lib/python3.14/site-packages/click/decorators.py", line 34, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "/home/toms/repos/GH/opensuse/docbuild.git/main/src/docbuild/cli/cmd_config/list.py", line 59, in list_config
    _print_section("Environment Configuration", context.envconfig.model_dump(), "env", flat, "yellow")
    ~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/toms/repos/GH/opensuse/docbuild.git/main/src/docbuild/cli/cmd_config/list.py", line 29, in _print_section
    print_json(data=data)
    ~~~~~~~~~~^^^^^^^^^^^
  File "/home/toms/repos/GH/opensuse/docbuild.git/main/.venv/lib/python3.14/site-packages/rich/__init__.py", line 106, in print_json
    get_console().print_json(
    ~~~~~~~~~~~~~~~~~~~~~~~~^
        json,
        ^^^^^
    ...<8 lines>...
        sort_keys=sort_keys,
        ^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/home/toms/repos/GH/opensuse/docbuild.git/main/.venv/lib/python3.14/site-packages/rich/console.py", line 1790, in print_json
    json_renderable = JSON.from_data(
        data,
    ...<7 lines>...
        sort_keys=sort_keys,
    )
  File "/home/toms/repos/GH/opensuse/docbuild.git/main/.venv/lib/python3.14/site-packages/rich/json.py", line 85, in from_data
    json = dumps(
        data,
    ...<6 lines>...
        sort_keys=sort_keys,
    )
  File "/home/toms/.local/share/uv/python/cpython-3.14.3-linux-x86_64-gnu/lib/python3.14/json/__init__.py", line 242, in dumps
    **kw).encode(obj)
          ~~~~~~^^^^^
  File "/home/toms/.local/share/uv/python/cpython-3.14.3-linux-x86_64-gnu/lib/python3.14/json/encoder.py", line 202, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/home/toms/.local/share/uv/python/cpython-3.14.3-linux-x86_64-gnu/lib/python3.14/json/encoder.py", line 263, in iterencode
    return _iterencode(o, 0)
  File "/home/toms/.local/share/uv/python/cpython-3.14.3-linux-x86_64-gnu/lib/python3.14/json/encoder.py", line 182, in default
    raise TypeError(f'Object of type {o.__class__.__name__} '
                    f'is not JSON serializable')
TypeError: Object of type IPv4Address is not JSON serializable
when serializing dict item 'host'
when serializing dict item 'server'
```
</details>

It looks like it's a problem with our models?

src/docbuild/cli/cmd_config/validate.py
```diff
@@ -0,0 +1,57 @@
+"""CLI interface to validate the configuration files."""
+
+import click
+from rich.console import Console
+from rich.panel import Panel
+
+
+@click.command(name="validate")
+@click.pass_context
+def validate_config(ctx: click.Context) -> None:
```
tomschr: This will validate _all_ config files. I think, that's unfortunate; maybe you only want to validate the XML? Or the env config?

I would introduce the options `--portal`, `--env`, and `--app` to restrict it to the specific config only. Additionally, the code below also introduces an `--all` option. The idea would be that `--all` and the other three options are mutually exclusive. 

This how you could add more arguments with an opt-in logic:

```suggestion
@click.command(name="validate")
@click.option('portal_flag',
    '--portal/--no-portal',
    default=False,
    help="Validate the portal settings"
)
@click.option('env_flag',
    '--env/--no-env',
    default=False,
    help="Validate the environment settings"
)
@click.option('app_flag',
    '--app/--no-app',
    default=False,
    help="Validate the app settings"
)
@click.option('all_flag',
    '--all',
    is_flag=True,
    default=True,
    help="Validate everything (Exclusive)."
)
@click.pass_context
def validate_config(ctx: click.Context, portal_flag, env_flag, app_flag, all_flag) -> None:
```

With that options, you would have

tomschr (changes requested):
Thanks for all your efforts! 👍
I realized that I didn't tell you an important aspect. I've described it below.

Have added some suggestions around some design questions.

src/docbuild/cli/cmd_config/validate.py
sushant-suse: Ah okay, thanks Toms. I have updated the things accordingly.

src/docbuild/cli/cmd_cli.py
sushant-suse: Thanks Toms, I agree. have modified the things accordingly.

src/docbuild/cli/cmd_config/list.py
sushant-suse: Agree. Done.

sushant-suse: Great catch Toms. Modified.

src/docbuild/cli/cmd_config/list.py
sushant-suse: Thanks for the sharp catch on the `IPv4Address` traceback and the suggestions for `flatten_dict` Toms. I have implemented the changes.

src/docbuild/cli/cmd_config/validate.py
sushant-suse: Great idea. Thanks

src/docbuild/cli/cmd_config/validate/__init__.py
```diff
@@ -0,0 +1,106 @@
...
+import click
+from rich.console import Console
+from rich.panel import Panel
+
+from .app import validate_app_settings
+from .env import validate_env_settings
+from .portal import validate_portal_content
+
+
+@click.command(name="validate")
```
tomschr: When you use `docbuild config validate -h`, Click outputs the docstring below. However, that contains the `:param:` descriptions of the function argument. I guess, we don't want that as it looks a bit ugly. 😆

Therefore, I'd like to suggest to amend this decorator:


```suggestion
@click.command(name="validate", help="Validate configuration files and portal content")
```

src/docbuild/cli/cmd_config/list.py
```diff
@@ -0,0 +1,77 @@
...
+    :param title: The title of the section to print.
+    :param data: The configuration data to print (as a dictionary).
+    :param prefix: The prefix to use for flat keys (e.g., "app" or "env").
+    :param flat: Whether to print in flat format or as JSON.
+    :param color: The color to use for the keys in flat format.
+    """
+    if flat:
+        # Use the generator directly to avoid holding a full list in memory
+        for k, v in flatten_dict(data, prefix):
+            console.print(f"[bold {color}]{k}[/bold {color}] = [green]{v}[/green]")
```
tomschr: When I use the `--flat` option, I get something like this:

```ini
app.logging.version = 1
app.logging.disable_existing_loggers = False
app.logging.formatters.standard.format = [%(asctime)s] [%(levelname)s] %(name)s: %(message)s
app.logging.formatters.standard.datefmt = %Y-%m-%d %H:%M:%S
app.logging.formatters.standard.style = None
[...]
```

That's fine. However, I would suggest a small addition:


```suggestion
            console.print(f"[bold {color}]{k}[/bold {color}] = [green]{repr(v)}[/green]")
```

This little addition makes the following output:

```ini
app.logging.version = 1
app.logging.disable_existing_loggers = False
app.logging.formatters.standard.format = '[%(asctime)s] [%(levelname)s] %(name)s: %(message)s'
app.logging.formatters.standard.datefmt = '%Y-%m-%d %H:%M:%S'
app.logging.formatters.standard.style = None
[...]
```

Basically it quote strings, but keep other basic datatypes untouched. That is very important when you have an empty string as value. The first case would just print a space, the second prints `''`.

src/docbuild/cli/cmd_config/list.py
```diff
@@ -0,0 +1,77 @@
...
+    for doctype in doctypes:
+        name = getattr(doctype, "name", "Unknown")
+        path = str(getattr(doctype, "path", "N/A"))
+        if flat:
+            console.print(f"[bold magenta]portal.{name}[/bold magenta] = [green]{path}[/green]")
+        else:
+            console.print(f"  - [bold]{name}[/bold]: {path}")
+
+
+@click.command(name="list")
```
tomschr, line 24: Add also a `help` argument in the decorator to avoid the ugly `:param` output issue:


```suggestion
@click.command(name="list", help="List the configuration as JSON or flat text.")
```

src/docbuild/utils/flatten.py
tomschr, line 1: I think you don't have a test case for this function. Shouldn't be too difficult to add that, right? 😉

src/docbuild/cli/cmd_config/validate/portal.py
tomschr, line 1: Sorry that I realized it too late. Could you remove all changes that belongs to XML validation?

tomschr (changes requested):
Thank you very much Sushant! Great work! 👍

I'm sorry that I found some issues that I wasn't aware before. Especially the validation part is tricky.

The validation of Docserv config works differently between old and new. With the new Portal schema, validation becomes actually easier.

As I'm still fixing the last bits on PR #197, I think we should remove the validation part for XML for the time being. Therefor it doesn't make sense to implement XML validation in this PR. It's also easier for *you* and you don't have to touch that twice.

For this reason, I've created issue #231 to separate these concerns. You can find all the details there. I also attached #231 as sub-issue.

Let me know what you think and sorry for the hassle.

tomschr (changes requested):
I just realized, we also need to adjust the documentation:

* https://opensuse.github.io/docbuild/user/config/index.html

  Mention the additional options and adjust the commands. See the files in `docs/source/user/config`.
  
* https://opensuse.github.io/docbuild/user/validate.html

  This needs to be removed as we will have a different subcommand in the future. See file `docs/source/user/validate.rst`

src/docbuild/cli/cmd_config/validate/__init__.py
sushant-suse: Agree Agree. Updated this one.

src/docbuild/cli/cmd_config/list.py
sushant-suse: Thanks Toms, this is nice addition.

src/docbuild/utils/flatten.py
sushant-suse: Definitely, added it.

src/docbuild/cli/cmd_config/validate/portal.py
sushant-suse: Sure, got it

sushant-suse:
> Thank you very much Sushant! Great work! 👍
> 
> I'm sorry that I found some issues that I wasn't aware before. Especially the validation part is tricky.
> 
> The validation of Docserv config works differently between old and new. With the new Portal schema, validation becomes actually easier.
> 
> As I'm still fixing the last bits on PR #197, I think we should remove the validation part for XML for the time being. Therefor it doesn't make sense to implement XML validation in this PR. It's also easier for _you_ and you don't have to touch that twice.
> 
> For this reason, I've created issue #231 to separate these concerns. You can find all the details there. I also attached #231 as sub-issue.
> 
> Let me know what you think and sorry for the hassle.



> Thank you very much Sushant! Great work! 👍
> 
> I'm sorry that I found some issues that I wasn't aware before. Especially the validation part is tricky.
> 
> The validation of Docserv config works differently between old and new. With the new Portal schema, validation becomes actually easier.
> 
> As I'm still fixing the last bits on PR #197, I think we should remove the validation part for XML for the time being. Therefor it doesn't make sense to implement XML validation in this PR. It's also easier for _you_ and you don't have to touch that twice.
> 
> For this reason, I've created issue #231 to separate these concerns. You can find all the details there. I also attached #231 as sub-issue.
> 
> Let me know what you think and sorry for the hassle.

Hi Toms, I've updated the PR as requested.

I have removed all logic related to XML/Portal validation and moved it to a 'deferred' status. The validate command now focuses strictly on TOML health. I've also kept the core fixes for the JSON serialization bug and added the requested unit tests for the flattening utility.

Let me know if this streamlined version is better.

src/docbuild/cli/cmd_validate/__init__.py
tomschr, line 1: Shouldn't this file be unmodified? That should be touch in a different PR.

src/docbuild/cli/cmd_validate/process.py
tomschr, line 1: Also this one (should be touched in a different PR.)

tests/cli/cmd_validate/validate/test_cmd_validate.py
tomschr, line 1: Shouldn't this file be unmodified? That should be touch in a different PR.

tests/cli/cmd_validate/validate/test_process_validation.py
tomschr, line 1: Shouldn't this file be unmodified? That should be touch in a different PR.

tests/cli/cmd_validate/test_process.py
tomschr, line 1: Shouldn't this file be unmodified? That should be touch in a different PR.

docs/source/user/config/index.rst
```diff
@@ -1,29 +1,26 @@
-Showing Configuration
-=====================
+Viewing and Validating Configuration
+------------------------------------
 
-The docbuild tool distinguishes between two types of configuration files:
+You can use the :command:`config` subcommand to list or validate your current settings.
```
tomschr:
```suggestion
Use the :command:`config` subcommand to list or validate your current settings.
```

docs/source/user/config/index.rst
```diff
@@ -1,29 +1,26 @@
...
 
-Additionally, the docbuild tool has also hardcoded default values for both types of configuration.
+.. code-block:: shell
 
-If no env or app configuration files are found, the docbuild tool will
-fallback to these hardcoded default values.
+   docbuild config list
 
-Keep in mind that these defaults may not be suitable for all use cases, and it is recommended to create and use a configuration file to customize the behavior of the docbuild tool.
+You can use the ``--flat`` flag to see the dotted-path format, or filter by ``--app`` or ``--env``.
```
tomschr:
```suggestion
Use the ``--flat`` flag to see the dotted-path format, or filter by ``--app`` or ``--env``.
```

src/docbuild/cli/cmd_config/__init__.py
tomschr, line 1: When I call `docbuild config -h`, I get the following validation error (see below). Also with `docbuild config list -h`. 

Shouldn't that be a lazy validation? Meaning it won't be triggered when user uses `-h`/`--help`.

```
$ docbuild config -h
15 Validation errors in config file 'unknown':

(1) In 'paths.repo_dir':
    Value error, Cannot create directory '/data/docserv/repos/permanent-full'. Permission denied: Parent directory '/data' is not writable.
    Expected: Permanent Repository Directory
    See: https://opensuse.github.io/docbuild/latest/errors/value_error.html

(2) In 'paths.tmp_repo_dir':
    Field required
    Expected: Temporary Repository Directory
    See: https://opensuse.github.io/docbuild/latest/errors/missing.html

(3) In 'paths.base_cache_dir':
    Value error, Cannot create directory '/var/cache/docserv'. Permission denied: Parent directory '/var/cache' is not writable.
    Expected: Base Cache Directory
    See: https://opensuse.github.io/docbuild/latest/errors/value_error.html

(4) In 'paths.base_server_cache_dir':
    Value error, Cannot create directory '/var/cache/docserv/doc-example-com'. Permission denied: Parent directory '/var/cache' is not writable.
    Expected: Base Server Cache Directory
    See: https://opensuse.github.io/docbuild/latest/errors/value_error.html

(5) In 'paths.meta_cache_dir':
    Value error, Cannot create directory '/var/cache/docserv/doc-example-com/meta'. Permission denied: Parent directory '/var/cache' is not writable.
    Expected: Metadata Cache Directory
    See: https://opensuse.github.io/docbuild/latest/errors/value_error.html

... and 10 more errors. Use '-vv' to see all errors.
```

src/docbuild/cli/cmd_config/list.py
tomschr, line 1: I think I found a regression. 🤔

When I use in the key `server.role` the value `devel_` (with an underscore!) I get this error message:

```
$ docbuild --env-config=env.devel.toml config list
1 Validation error in config file 'env.devel.toml':

(1) In 'server.role':
    Input should be 'production', 'staging', 'testing', 'production', 'production', 'production', 'production', 'staging', 'staging', 'staging', 
'staging', 'testing', 'testing', 'testing', 'testing', 'testing', 'testing', 'testing' or 'testing'
    Expected: Server Role
    See: https://opensuse.github.io/docbuild/latest/errors/enum.html
```

I think we had it fixed some time ago, right? Maybe something for a different PR. 🙂

tomschr (changes requested):
Thank you very much, Sushant! Great work! 👍

I have only two issue: 

* Could you try to revert the deletion of anything in regards to XML validation (code + test)? We need that in a later, follow-up PR.
  Or do you see a better solution?
* Lazy validation? The script doesn't show the help output but tries to validate it. Not sure if this could be caught, but that's a bit unfortunate. IMHO, the script should trigger validation when the user tries to show the help page.

src/docbuild/cli/cmd_config/__init__.py
sushant-suse: Yes, fixing it.

src/docbuild/cli/cmd_config/list.py
sushant-suse: Strange let me check.

tomschr (changes requested):
All fine now, but docs in regards to env/app aren't visible, right?

sushant-suse:
> All fine now, but docs in regards to env/app aren't visible, right?

Hi Toms, added them back

tomschr (approved):
Thank you very much! 👍

PR #251, opened 2026-05-04
fix #232: implement PEP 517 build wrapper for automatic RNC to RNG conversion

### LINKED ISSUE #232: Investigate a PEP 517 build wrapper for RNC -> RNG
### Situation

Our project uses the declarative `pyproject.toml` approach. That gives us many advantages. However, it also limits _how_ to build when it's needed.

This is especially important when we use our Portal RELAX NG schema. It's developed in the compact form, but for technically reasons, we need both formats (RNC and RNG). As such, only the RNC format is versioned by Git, but RNG should be converted on demand.

### Use Case

Both formats are equivalent, but not all tools support both. Some need the RNG (XML) format:

* `lxml` for validation
* Sphinx extensions that tries to create documentation.

### Possible Implementation

The idea is to use an in-tree build backend method that follows [PEP 517](https://peps.python.org/pep-0517/). We create a wrapper script that does:

1. Run `trang` go generate the `.rng` file.
2. Hand over the rest of the heavy lifting to the standard `setuptools.build_meta` functions.

The build script `_docbuild_backend.py` would look like this:

```python
# file _docbuild_backend.py
import subprocess
import os
from setuptools import build_meta as _orig

# Proxy all required PEP 517 hooks to setuptools.build_meta
prepare_metadata_for_build_wheel = _orig.prepare_metadata_for_build_wheel
get_requires_for_build_wheel = _orig.get_requires_for_build_wheel
get_requires_for_build_sdist = _orig.get_requires_for_build_sdist

def run_trang():
    """Helper to run the conversion."""
    rnc_path = os.path.join("mypackage", "portal-config-schema.rnc")
    rng_path = os.path.join("mypackage", "portal-config-schema.rng")
    
    if os.path.exists(rnc_path):
        print(f"--- Custom Build Hook: Converting {rnc_path} to {rng_path} ---")
        subprocess.run(["trang", rnc_path, rng_path], check=True)

def build_wheel(wheel_directory, config_settings=None, metadata_directory=None):
    run_trang()
    return _orig.build_wheel(wheel_directory, config_settings, metadata_directory)

def build_sdist(sdist_directory, config_settings=None):
    run_trang()
    return _orig.build_sdist(sdist_directory, config_settings)
```

Now, we need to tell our project to use this local script as the build engine:

```toml
[build-system]
# Include setuptools and wheel as requirements for your backend to work
requires = ["setuptools", "wheel"]
# Use your local file (without the .py extension) as the backend
build-backend = "_docbuild_backend"
# Tell Python to look in the current directory ('.') for the backend script
backend-path = ["."]

# Rest the same...
```

You can build it by using either of the following methods:

* Standard "pip":

  ```shell
  $ pip install build
  $ python -m build
  ```

* Using "uv":

  ```shell
  $ uv build
  ```

sushant-suse:
Fixes #232

**What changed**:
Implemented an in-tree PEP 517 build backend (`_docbuild_backend.py`) to automatically convert RELAX NG Compact (`.rnc`) schemas into XML (`.rng`) format prior to packaging. 

**How it works**:
* Created `_docbuild_backend.py` to intercept the `build_wheel` and `build_sdist` hooks.
* The wrapper runs `trang` to compile any `.rnc` files in `src/docbuild/config/xml/data/` into `.rng`.
* Updated `pyproject.toml` to use the local build backend and include `.rng` files in `package-data`.
* Added `MANIFEST.in` to ensure the custom build script is included in source distributions.


_docbuild_backend.py
```diff
@@ -0,0 +1,85 @@
...
+        return
+
+    # Find all .rnc files in the directory
+    for rnc_path in data_dir.glob("*.rnc"):
+        rng_path = rnc_path.with_suffix(".rng")
+
+        print(f"--- Custom Build Hook: Converting {rnc_path} to {rng_path} ---")
+        try:
+            # Run trang command
+            subprocess.run(["trang", str(rnc_path), str(rng_path)], check=True)
```
tomschr, line 35: Maybe add a line to inform the RNC -> RNG was successful?


```suggestion
            subprocess.run(["trang", str(rnc_path), str(rng_path)], check=True)
            print(f"Successfully converted {rnc_path} -> {rng_path}")
```

_docbuild_backend.py
```diff
@@ -0,0 +1,85 @@
+"""Custom PEP 517 build backend to convert RNC schemas to RNG before packaging."""
+
+from pathlib import Path
+import subprocess
+from typing import Any
```
tomschr: Import `sys`:


```suggestion
from typing import Any
import sys
```

_docbuild_backend.py
```diff
@@ -0,0 +1,85 @@
...
+        try:
+            # Run trang command
+            subprocess.run(["trang", str(rnc_path), str(rng_path)], check=True)
+        except FileNotFoundError:
+            print(
+                "WARNING: 'trang' command not found. "
+                "Ensure it is installed to generate RNG schemas."
+            )
+        except subprocess.CalledProcessError as e:
+            print(f"ERROR: trang conversion failed for {rnc_path}")
```
tomschr: Those error message should go to `sys.stderr`:


```suggestion
            print(
                "WARNING: 'trang' command not found. "
                "Ensure it is installed to generate RNG schemas.",
                file=sys.stderr
            )
        except subprocess.CalledProcessError as e:
            print(f"ERROR: trang conversion failed for {rnc_path}", file=sys.stderr)
```

src/docbuild/config/xml/data/portal-config.rng
tomschr, line 1: This file is generated and shouldn't be versioned by git. I think it's better to remove that.

changelog.d/232.feature.rst
```diff
@@ -0,0 +1 @@
+Added an in-tree PEP 517 build backend to automatically compile RELAX NG Compact (``.rnc``) schemas into XML (``.rng``) format during the build process, ensuring both formats are available in the distributed package.
```
tomschr: Strictly speaking, we have only one schema (`portal-config.rnc`), which is the Portal schema. Perhaps we will have variants (a stricter or more relaxed) in the future. But for the time being, I'd suggest:


```suggestion
Added an in-tree PEP 517 build backend to automatically compile the Portal Schema from  RELAX NG Compact (``.rnc``) into XML (``.rng``) format during the build process, ensuring both formats are available in the distributed package.
```

tomschr (changes requested):
That's a great implementation! I've tested to build wheel and `.tar.gz` and both work! 👍

I have some minor comments, see below.

changelog.d/232.feature.rst
sushant-suse: Thanks Toms, agree.

src/docbuild/config/xml/data/portal-config.rng
sushant-suse: Removed it

tomschr (approved):
Great work, thanks a lot Sushant! 👍

PR #256, opened 2026-05-05
Add convenience wrapper script for Portal

tomschr:
Add two convenience wrapper scripts:

* `migrate.sh`
  Converts an old docserv-stitchfile.xml into the new Portal config.
* `validate.sh`
  Validates the new Portal file (and everything that was included).

### DIFF
--- a/changelog.d/256.feature.rst
+++ b/changelog.d/256.feature.rst
@@ -1,0 +1,1 @@
+Add convenience wrapper scripts for Portal validation and migration from old Docserv config.
--- a/src/docbuild/config/xml/data/migrate.sh
+++ b/src/docbuild/config/xml/data/migrate.sh
@@ -17,6 +17,13 @@ XSLT="$SCRIPT_DIR/convert-v6-to-v7.xsl"
 INPUT="docserv-stitch-2026-04-27.xml"
 USE_XINCLUDE=false
 
+
+if ! command -v xsltproc >/dev/null 2>&1; then
+    echo "Error: 'xsltproc' is required but not installed." >&2
+    exit 1
+fi
+
+
 # --- Help Function ---
 usage() {
     # Using shell parameter expansion ${0##*/} instead of basename $0
@@ -80,7 +87,7 @@ if [ "$USE_XINCLUDE" = true ]; then
     # --stringparam use.xincludes 1 passes the parameter to your XSLT logic
     # Note: You may also need to add the physical --xinclude flag here if xsltproc
     # needs to resolve the tags before passing them to the XSLT.
-    XINCLUDE_FLAGS="--stringparam use.xincludes 1"
+    XINCLUDE_FLAGS="--xinclude --stringparam use.xincludes 1"
 fi
 
 # --- Execution ---
--- a/src/docbuild/config/xml/data/migrate.sh
+++ b/src/docbuild/config/xml/data/migrate.sh
@@ -1,0 +1,92 @@
+#!/bin/sh
+
+# --- Default Values ---
+SCHEMAFILE="portal-config.rnc"
+OUTPUT="portal-test.xml"
+OUTDIR="./"
+
+# Get the directory where the script is located using shell built-ins
+# If $0 contains a slash, SCRIPT_DIR is the part before the last slash.
+# Otherwise, we assume the current directory (.).
+case "$0" in
+    */*) SCRIPT_DIR="${0%/*}" ;;
+    *)   SCRIPT_DIR="." ;;
+esac
+
+XSLT="$SCRIPT_DIR/convert-v6-to-v7.xsl"
+INPUT="docserv-stitch-2026-04-27.xml"
+USE_XINCLUDE=false
+
+# --- Help Function ---
+usage() {
+    # Using shell parameter expansion ${0##*/} instead of basename $0
+    SCRIPT_NAME=${0##*/}
+    cat << EOF
+Usage: $SCRIPT_NAME [OPTIONS] [INPUT_FILE]
+
+Options:
+  -s, --schema FILE      Path to schema file (Default: $SCHEMAFILE)
+  -o, --output FILE      Name of the output file (Default: $OUTPUT)
+  -d, --dir DIR          Output directory (Default: $OUTDIR)
+  -x, --xinclude         Enable xinclude processing
+  -h, --help             Show this help message
+
+Arguments:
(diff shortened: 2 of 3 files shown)


sushant-suse:
> Add two convenience wrapper scripts:
> 
> * `migrate.sh`
>   Converts an old docserv-stitchfile.xml into the new Portal config.
> * `validate.sh`
>   Validates the new Portal file (and everything that was included).

Hey Toms, this is an incredibly clean and well-structured POSIX shell script. Love the manual argument parsing and the case fallback for `SCRIPT_DIR`. 

I spotted a couple of minor things we can tighten up to make it better (only if you find them really helpful):

#### 1. Physical `--xinclude` for `xsltproc`

If the `INPUT` file relies on physical XML inclusions, `xsltproc` needs its own `--xinclude` CLI flag to process them. Right now, `-x` only passes `--stringparam use.xincludes 1` to the stylesheet. 

We should probably append the literal `--xinclude` flag to the execution command when `USE_XINCLUDE` is true:

```sh
XINCLUDE_FLAGS=""
if [ "$USE_XINCLUDE" = true ]; then
    # Enable it for the stylesheet AND the parser itself
    XINCLUDE_FLAGS="--xinclude --stringparam use.xincludes 1"
fi
```

WDYT?

#### 2. Quote protection for paths containing spaces In the final execution step.

Wrapping `$XSLT` and `$INPUT` in double quotes will prevent the script from crashing if someone executes it within a path containing spaces:

```sh
xsltproc $XINCLUDE_FLAGS \
         --stringparam schemafile "$SCHEMAFILE" \
         --stringparam outputfile "$OUTPUT" \
         --stringparam outputdir "$OUTDIR" \
         "$XSLT" "$INPUT"
```

#### 3. Fail-early dependency check (Optional but nice to have)

Just in case someone runs this in a stripped-down container environment, we could add a quick check at the top to ensure `xsltproc` is available:
```sh
if ! command -v xsltproc >/dev/null 2>&1; then
    echo "Error: 'xsltproc' is required but not installed." >&2
    exit 1
    fi
```

---
 
For the `validate.sh`, again excellent work. I found a couple of areas where we can make it completely robust across all shells and operating systems:
 
#### 1. Fixing the XInclude Java Flags

I noticed we defined `$XINCLUDE_PROP` but passed it as `ADDITIONAL_FLAGS` to the command execution:

```sh
ADDITIONAL_FLAGS="$XINCLUDE_PROP" jing ...
```
On many systems, the `jing` shell wrapper won't recognize `ADDITIONAL_FLAGS`. To make sure the JVM picks up the parser configuration on both openSUSE/Fedora (`JAVA_OPTS`) and Debian/Ubuntu (`JAVA_ARGS`), we should assign them explicitly:

```sh
if [ "$USE_XINCLUDE" = true ]; then
    # Assign both wrapper variables directly before executing jing
    JAVA_OPTS="$XINCLUDE_PROP" JAVA_ARGS="$XINCLUDE_PROP" jing $JING_FLAGS "$SCHEMAFILE" "$INPUT"
else
    jing $JING_FLAGS "$SCHEMAFILE" "$INPUT"
fi
```

WDYT?
 
#### 2. POSIX-compliant Variable Concatenation

Since the shebang is `#!/bin/sh`, some environments (like Debian/Ubuntu using `dash`) will crash on the `+=` syntax:

```sh
JING_FLAGS+="-c "
```

We can make this fully POSIX-compliant by rewriting it using standard concatenation:
```sh
case "$SCHEMAFILE" in
    *.rnc)
        JING_FLAGS="$JING_FLAGS -c"
        ;;
esac

if [ "$DISABLE_ID_CHECK" = true ]; then
    JING_FLAGS="$JING_FLAGS -i"
fi
```

#### 3. Double-quoting `$JING_FLAGS`

When invoking `jing`, passing unquoted variables that might be empty can sometimes cause execution hiccups. However, if we quote `"$JING_FLAGS"`, it will pass an empty string `""` as an argument if no flags are set, which `jing` might complain about. A robust shell trick to expand it only if it has a value is using `${JING_FLAGS}` without quotes, or constructing it carefully. Let's make sure we double-quote the entry files:

```sh
jing $JING_FLAGS "$SCHEMAFILE" "$INPUT"
```

WDYT?

---

I don't know but I am not able to directly attach my suggestions in file (maybe some GitHub issue). Sorry for the inconvenience.

sushant-suse (approved):

tomschr:
Thank you for your valuable comments! Much appreciated! :+1: 

Here are some comments to your posts:

### `migrate.sh`

> 1. Physical --xinclude for xsltproc
> [...]
> We should probably append the literal --xinclude flag to the execution command when USE_XINCLUDE is true:

The `--xinclude` flag is needed when the _source_ file contains XInclude elements. When I pass a Docserv stitch file, this is not the case; it's one big file. Everything is already resolved. That's why I didn't include the flag.

Nevertheless, I think your suggestions is good and I included that.


> 2. Quote protection for paths containing spaces In the final execution step.
> [...]

You are correct, but the quotes are already there, aren't they?

> 3. Fail-early dependency check (Optional but nice to have)

That's a good idea! I use that for `xsltproc` and for `jing`.


### `validate.sh`

> 1. Fixing the XInclude Java Flags
> [...] On many systems, the jing shell wrapper won't recognize ADDITIONAL_FLAGS.

You are spot on! :+1: I'm aware of that, but as this is mostly(?) run on openSUSE, I was a bit lazy and didn't take that into account.

I know little about how Debian or Fedora create their Java wrapper scripts and how to influence them. Is it enough to just set the `JAVA_OPTS` variable? Or do we need to check which OS is used? How about MacOS?

I've tested both, the old implementation with `ADDITIONAL_FLAGS` and the `JAVA_OPTS`/`JAVA_ARGS`. This is the result:

* `ADDITIONAL_FLAGS`: works. The XInclude elements are resolved.
* `JAVA_OPTS` and `JAVA_ARGS`: don't work. It outputs an error which is clearly related to XInclude processing didn't happen (`error: element "xi:include" not allowed yet`)


> 2. POSIX-compliant Variable Concatenation
You are absolutely right to mention the `+=` syntax be not POSIX-compliant. I've fixed that.


> 3. Double-quoting $JING_FLAGS
Thanks, I wasn't aware of that. The `$JING_FLAGS` are already unquoted so it and the schema file and input are quoted.


I've tried to fix all in commit 8ab4e860.

sushant-suse (approved):

PR #279, opened 2026-05-15
infra: add data-driven test suite for Portal XML schema (#258)

### LINKED ISSUE #258: Add tests for the Portal Schema
### Situation

The new Portal schema is complex. It may be helpful to have some good and bad examples which could test some aspects of the schema.

### Use Case

Finding edge cases and have examples of good and bad practices.

#### Strategy

* Positve tests: Files that _must_ pass (Jing returns exit code 0)
* Negative tests: Files that _must fail_ (Jing returns a non-zero exit code).

#### Test hierarchy

It's recommended to have a directory structure similar like this:

```
tests/
│   ├── test_schema.py
│   ├── valid_cases/      <-- Drop "good" XMLs here
│   │   ├── basic.xml
│   │   └── complex.xml
│   └── invalid_cases/    <-- Drop "bad" XMLs here
│       ├── missing_attr.xml
│       └── malformed.xml
```

```python
import subprocess
import pytest
from pathlib import Path

# Configuration - Adjust these paths as needed
BASE_DIR = Path(__file__).parent
SCHEMA = BASE_DIR / ".../portal-config.rng"
VALID_DIR = BASE_DIR / "valid_cases"
INVALID_DIR = BASE_DIR / "invalid_cases"


def run_jing(xml_path):
    """
    Standard Linux subprocess call to Jing.
    Returns the exit code.
    """
    # Use 'sudo apt install jing' if not already present
    result = subprocess.run(
        ['jing', str(SCHEMA), str(xml_path)],
        capture_output=True,
        text=True
    )
    return result.returncode, result.stderr

def get_xml_files(directory):
    """Helper to find all .xml files in a directory."""
    return list(directory.glob("*.xml"))


### 1. Dynamic Positive Tests
@pytest.mark.parametrize("xml_file", get_xml_files(VALID_DIR), ids=lambda p: p.name)
def test_valid_files(xml_file):
    exit_code, stderr = run_jing(xml_file)
    assert exit_code == 0, f"Valid file {xml_file.name} failed to validate: {stderr}"

### 2. Dynamic Negative Tests
@pytest.mark.parametrize("xml_file", get_xml_files(INVALID_DIR), ids=lambda p: p.name)
def test_invalid_files(xml_file):
    exit_code, _ = run_jing(xml_file)
    # We expect failure (non-zero)
    assert exit_code != 0, f"Invalid file {xml_file.name} unexpectedly passed validation!"
```

#### Special Attention
XInclude processing should be enabled (not shown above).
See `src/docbuild/config/xml/data/validate.sh`. Maybe we should use this script instead of `jing` directly?

#### Alternatives
See tests in https://github.com/openSUSE/geekodoc

sushant-suse:
Closes #258 

This PR implements a robust, data-driven testing framework for the **Portal XML Schema** (`portal-config.rnc`). It ensures that the complex rules of the portal configuration are strictly enforced and provides a mechanism to catch regressions in both the schema and configuration files.

### Changes

* **Added `tests/config/xml/test_schema.py**`: A new test suite that dynamically discovers and validates XML cases using the `jing` validator. It includes full **XInclude** support and cross-platform compatibility for different Java/Jing wrappers.
* **Added `tests/config/xml/schema_cases/**`: Created a hierarchy of test data:
* `valid/`: Contains "Positive" tests (minimal valid portal configs).
* `invalid/`: Contains "Negative" tests (configs with invalid DC prefixes, illegal lifecycle values, etc.).
* **Added News Fragment**: Created `changelog.d/258.infra.rst` for automated release notes.

### DIFF
--- a/tests/config/xml/test_schema.py
+++ b/tests/config/xml/test_schema.py
@@ -1,0 +1,56 @@
+import shutil
+import subprocess
+from pathlib import Path
+
+import pytest
+
+# Configuration of paths
+BASE_DIR = Path(__file__).parent
+DATA_DIR = BASE_DIR.parents[2] / "src/docbuild/config/xml/data"
+SCHEMA = DATA_DIR / "portal-config.rnc"
+
+# Local test case directories
+CASES_DIR = BASE_DIR / "schema_cases"
+VALID_DIR = CASES_DIR / "valid"
+INVALID_DIR = CASES_DIR / "invalid"
+
+def run_jing(xml_path: Path) -> tuple[int, str]:
+    """Execute jing with full XInclude support enabled."""
+    jing_bin = shutil.which("jing")
+    if not jing_bin:
+        pytest.skip("jing executable not found in PATH")
+
+    xinclude_prop = "-Dorg.apache.xerces.xni.parser.XMLParserConfiguration=" \
+                    "org.apache.xerces.parsers.XIncludeParserConfiguration"
+    
+    env = {
+        "JAVA_OPTS": xinclude_prop,
+        "JAVA_ARGS": xinclude_prop,
+        "ADDITIONAL_FLAGS": xinclude_prop
+    }
+
+    cmd = [jing_bin]
+    if SCHEMA.suffix == ".rnc":
+        cmd.append("-c")
+    
+    cmd.extend([str(SCHEMA), str(xml_path)])
+
+    result = subprocess.run(cmd, capture_output=True, text=True, env=env)
+    
+    # Combine stdout and stderr for the error message
+    output = (result.stdout + result.stderr).strip()
+    return result.returncode, output
+
+
+@pytest.mark.parametrize("xml_file", list(VALID_DIR.glob("*.xml")), ids=lambda p: p.name)
+def test_portal_schema_valid(xml_file: Path) -> None:
+    """Positive tests: These files must validate successfully."""
+    exit_code, stderr = run_jing(xml_file)
+    assert exit_code == 0, f"Valid file {xml_file.name} failed validation:\n{stderr}"
+
+
+@pytest.mark.parametrize("xml_file", list(INVALID_DIR.glob("*.xml")), ids=lambda p: p.name)
+def test_portal_schema_invalid(xml_file: Path) -> None:
+    """Negative tests: These files must fail validation."""
+    exit_code, _ = run_jing(xml_file)
+    assert exit_code != 0, f"Invalid file {xml_file.name} unexpectedly passed validation!"
--- a/tests/config/xml/test_schema.py
+++ b/tests/config/xml/test_schema.py
@@ -1,6 +1,6 @@
+from pathlib import Path
 import shutil
 import subprocess
-from pathlib import Path
 
 import pytest
 
@@ -22,7 +22,7 @@ def run_jing(xml_path: Path) -> tuple[int, str]:
 
     xinclude_prop = "-Dorg.apache.xerces.xni.parser.XMLParserConfiguration=" \
                     "org.apache.xerces.parsers.XIncludeParserConfiguration"
-    
+
     env = {
         "JAVA_OPTS": xinclude_prop,
         "JAVA_ARGS": xinclude_prop,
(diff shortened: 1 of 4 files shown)


tests/config/xml/test_schema.py
```diff
@@ -0,0 +1,56 @@
+from pathlib import Path
+import shutil
+import subprocess
+
+import pytest
+
+# Configuration of paths
+BASE_DIR = Path(__file__).parent
+DATA_DIR = BASE_DIR.parents[2] / "src/docbuild/config/xml/data"
+SCHEMA = DATA_DIR / "portal-config.rnc"
```
tomschr: That is technically correct. Probably we can leave it as is, but I'm wondering if it would be more "Pythonic" to use the `importlib.resources` module? Something like this (you need to import it first):


```suggestion
SCHEMA = resources.files("docbuild.config.xml").joinpath(
    "data", "portal-config.rnc"
)
```

The reason why I suggest that is, that VSCode _may_ recognize that when your refactor code (rename directory names, moving files etc). A simple string could be left untouched. 🤷‍♂️ 

But maybe this is not really needed and just a fever dream. Take my suggestion with some caution here.

tests/config/xml/test_schema.py
```diff
@@ -0,0 +1,56 @@
...
+# Local test case directories
+CASES_DIR = BASE_DIR / "schema_cases"
+VALID_DIR = CASES_DIR / "valid"
+INVALID_DIR = CASES_DIR / "invalid"
+
+def run_jing(xml_path: Path) -> tuple[int, str]:
+    """Execute jing with full XInclude support enabled."""
+    jing_bin = shutil.which("jing")
+    if not jing_bin:
+        pytest.skip("jing executable not found in PATH")
```
tomschr: This one here is actually more important. Perhaps we could be more Pytest-idiomatic here, I completely forgot about this feature.😉

I'd suggest to use a [`pytestmark`](https://docs.pytest.org/en/stable/reference/reference.html#globalvar-pytestmark) decorator. This applies to the complete test module. If `jing` could be found, all tests will be executed. If not, all module level tests are skipped and you see it in the output in the skipped counter. If we want to add more tests in the future, we don't forget it. 😉

Here is how it looks like:


```suggestion
JING_BIN = shutil.which("jing")

# Skip ALL tests in this module if jing is not available.
# jing is the RELAX NG validator required to validate portal XML configurations.
pytestmark = pytest.mark.skipif(
    JING_BIN is None,
    reason="jing executable not found in PATH",
)


def run_jing(xml_path: Path) -> tuple[int, str]:
    """Execute jing with full XInclude support enabled."""
```

tests/config/xml/test_schema.py
```diff
@@ -0,0 +1,56 @@
...
+    xinclude_prop = "-Dorg.apache.xerces.xni.parser.XMLParserConfiguration=" \
+                    "org.apache.xerces.parsers.XIncludeParserConfiguration"
+
+    env = {
+        "JAVA_OPTS": xinclude_prop,
+        "JAVA_ARGS": xinclude_prop,
+        "ADDITIONAL_FLAGS": xinclude_prop
+    }
+
+    cmd = [jing_bin]
```
tomschr:
```suggestion
    cmd = [JING_BIN]
```

tomschr (changes requested):
That's a great start! This is what I had exactly in my mind!😆

I have some minor suggestions, see below.

The idea would be to add more XML files and sort them into positive or negative tests. There are still many structures that aren't covered. Each of the tests covers only a specific aspect or structure that needs to be tested.

However, that's something we can do in a later PR. The groundwork is done. Great work! 👍

sushant-suse:
> That's a great start! This is what I had exactly in my mind!😆
> 
> I have some minor suggestions, see below.
> 
> The idea would be to add more XML files and sort them into positive or negative tests. There are still many structures that aren't covered. Each of the tests covers only a specific aspect or structure that needs to be tested.
> 
> However, that's something we can do in a later PR. The groundwork is done. Great work! 👍

Thanks Toms, updating the things as per your suggestions.

tests/config/xml/test_schema.py
sushant-suse: Not a fever dream at all; this is a fantastic suggestion Toms. Shifting away from fragile relative path walking (`parents[2]`) to `importlib.resources` is much cleaner and ensures the test won't break if we ever reorganize the test directory or package the tool. I will absolutely adopt this. Thanks.

tests/config/xml/test_schema.py
sushant-suse: Oh, I love this. Promoted the check to `pytestmark` at the module level. It's much cleaner to skip the entire suite declaratively up front rather than handling it dynamically inside the runner function. Great catch, thank you Toms.

tomschr (approved):
THank you very much! Just squash-merge it and you are done. 👍

PR #305, opened 2026-05-22
Use uv.lock file to prevent attacks

tomschr:
The dependency cooldown feature with `exclude-newer` seems to have some negative side effects for Dependabot. The workflow fails as uv can fullfil the dependencies anymore.

Removing all negative side effects from `uv.toml`. The committed `uv.lock` (with SHA-256 hashes) is the primary supply-chain protection. See: https://docs.astral.sh/uv/concepts/resolution/#exclude-newer

Related to PR #287.


sushant-suse (approved):

tomschr:
@copilot Add a news fragment file for Towncrier in the directory `changelog.d/`. Use the `changelog.d/README.rst` for details.

PR #319, opened 2026-05-28
feat: auto-complete partial language codes in LanguageCode

### LINKED ISSUE #309: Refactor LanguageCode: allow partial languages
### Situation

Sometimes deliverables contain partial languages like "en" instead of the full language code "en-us".

### Use Case

It would be helpful to be more lenient to accept even these partial language codes.

### Possible Implementation

As we don't have any ambiguous language code (for example, `en-us` vs. `en-uk`), we could use:
 
* LanguageCode could compare the first part with the `ALLOWED_LANGUAGES` constant and complete it to LANGUAGE-COUNTRY. For example, `en` would be rewritten to `en-us`.
* After whitespace handling, validate the data.
* Anything else that didn't match this still throws a `ValidateError`.

```python
>>> LanguageCode(language="en")
LanguageCode(language='en-us')
>>> LanguageCode(language=" de")
LanguageCode(language='de-de')
```

sushant-suse:
Closes #309 

### Changes

* Updated the `_normalize_language_separator` pre-validator in `LanguageCode` to detect partial languages (strings without a hyphen).
* If a partial language is detected, it searches `ALLOWED_LANGUAGES` and auto-completes to the full `LANGUAGE-COUNTRY` format.
* Retains standard `ValidationError` behavior for completely invalid inputs (e.g., `"fr"`).

### DIFF
--- a/changelog.d/259.bugfix.rst
+++ b/changelog.d/259.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed a validation bug in `LanguageCode` by automatically stripping leading and trailing whitespaces from input strings.
--- a/changelog.d/309.feature.rst
+++ b/changelog.d/309.feature.rst
@@ -1,0 +1,1 @@
+Refactored `LanguageCode` to automatically strip leading and trailing whitespaces and auto-complete partial language codes (for example, resolving `en` to `en-us`).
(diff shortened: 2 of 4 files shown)


tomschr (changes requested):
Thank you for your contribution! 👍

The code is excellent, but two things are missing:

* The newsfragment file.😉 
* Tests

tomschr (approved):
Perfect! Great work! 🎉 Just squash-merge it.

PR #348, opened 2026-06-16
cli: dynamically validate docsets and improve error output

### LINKED ISSUE #303: Use a list of available docsets when `portal list` fails
### Project Version

0.19.5

### Python Version

3.14

### Operating System

Linux

### Steps to Reproduce

Run the `docbuild` script with a invalid docset, for example `cloudnative/x`.

### Actual Behavior

```
$ docbuild --env-config env.devel.toml portal list cloudnative/x
[2026-05-20 10:54:48] [WARNING] docbuild.config.xml: No deliverables found for Doctype(product='cloudnative', docset=[x], lifecycle='unknown', langs=[en-us])
```

### Expected Behavior

It would be nice if `docbuild` could give you a list of available docsets. For example, if you make a typo, you get this:

```
$ docbuild --env-config env.devel.toml portal list slex
Error parsing doctype: 1 validation error for Doctype
product
  Value error, 'slex' is not a valid Product. Allowed values are: '*', 'appliance', 'cloudnative', 'compliance', 'container',
'liberty', 'sbp', 'ses', 'sled', 'sle-ha', 'sle-hpc', 'sle-micro', 'sle-public-cloud', 'sle-rt', 'sles-sap', 'sles', 
'sle-vmdp', 'smart', 'smt', 'soc', 'style', 'subscription', 'suma-retail', 'suma', 'suse-ai', 'suse-caasp', 'suse-cap', 
'suse-distribution-migration-system', 'suse-edge', 'trd' 
    For further information visit https://errors.pydantic.dev/2.13/v/value_error
Aborted!
```

IMHO, if you make a typo for a docset, it should be consistent with a product.

### Additional Information

_No response_

sushant-suse:
Closes #303

Implemented dynamic validation for docsets in the `portal list` command. Previously, entering an invalid docset (like `cloudnative/x`) would silently fail and return a generic "0 deliverables" warning. Now, it intercepts the typo, queries the configuration for valid docsets for that specific product, and raises a helpful error message.

### Changes:

* **`src/docbuild/cli/cmd_portal/cmd_list.py`**: Added `validate_docsets_against_xml` to dynamically cross-reference parsed doctypes against the loaded XML tree.
* Injected this validation step inside `async_list_cmd` right after the XML is parsed but before the deliverables are queried.
* Formatted the resulting `click.UsageError` to exactly match the styling of Pydantic's static validation errors for consistency.
* Added news fragment `303.bugfix.rst`.

### Self-Review Checklist

- [x] Code is well-structured and easy to follow.
- [x] Dynamic validation correctly catches typos without breaking existing fallback/wildcard logic.
- [x] Error message format matches the expected Pydantic standard requested in the issue.

**Documentation & CI**:

- [N/A] Does the issue need a documentation update?

### DIFF
--- a/changelog.d/303.bugfix.rst
+++ b/changelog.d/303.bugfix.rst
@@ -1,0 +1,1 @@
+Added dynamic validation to the ``portal list`` command to provide helpful error messages with available allowed values when an invalid docset is provided.
--- a/changelog.d/303.bugfix.rst
+++ b/changelog.d/303.bugfix.rst
@@ -1,1 +1,1 @@
-Added dynamic validation to the ``portal list`` command to provide helpful error messages with available allowed values when an invalid docset is provided.
+Added dynamic validation to the :command:`portal list` command to provide helpful error messages with available allowed values when an invalid docset is provided.
--- a/src/docbuild/cli/cmd_portal/cmd_list.py
+++ b/src/docbuild/cli/cmd_portal/cmd_list.py
@@ -181,7 +181,7 @@ def validate_docsets_against_xml(
         if dt.product and "*" not in dt.product and dt.docset and "*" not in dt.docset:
             prod_val = dt.product.value
             # Fetch all valid 'path' attributes from docsets under this specific product
-            valid_docsets = tree.xpath(f"//product[@id='{prod_val}']/docset/@path")
+            valid_docsets = tree.xpath(f"//product[@id={prod_val!r}]/docset/@path")
 
             # Format them for display, always prepending the wildcard allowed value
             valid_docsets_str = ["*", *sorted(str(v) for v in valid_docsets)]
@@ -189,7 +189,7 @@ def validate_docsets_against_xml(
             for ds in dt.docset:
                 if ds not in valid_docsets:
                     allowed_str = ", ".join(f"'{v}'" for v in valid_docsets_str)
-                    console.print("[red]Error parsing doctype: 1 validation error for Doctype[/red]")
+                    console.print("[red]Error parsing doctype:[/red] 1 validation error for Doctype")
                     console.print("docset")
                     console.print(f"  Value error, '{ds}' is not a valid Docset. Allowed values are: {allowed_str}")
                     raise click.Abort()
--- a/src/docbuild/cli/cmd_portal/cmd_list.py
+++ b/src/docbuild/cli/cmd_portal/cmd_list.py
@@ -181,10 +181,10 @@ def validate_docsets_against_xml(
     for dt in doctypes:
         if dt.product and "*" not in dt.product and dt.docset and "*" not in dt.docset:
             prod_val = dt.product.value
-            
+
             # Use the class's own string parser to bypass strict __init__ type-checking issues
             broad_dt = Doctype.from_str(f"{prod_val}/*/*")
-            
+
             # Harvest all valid docset IDs using the Deliverable abstraction
             valid_docsets = set()
             for node in list_all_deliverables(tree, [broad_dt]):
--- a/src/docbuild/cli/cmd_portal/cmd_list.py
+++ b/src/docbuild/cli/cmd_portal/cmd_list.py
@@ -172,6 +172,29 @@ def print_hierarchy(
         console.print()
 
 
+def validate_docsets_against_xml(
+    doctypes: list[Doctype], tree: etree._ElementTree | etree._Element, console: Console
+) -> None:
+    """Dynamically validate that provided docsets exist for their respective products."""
+    for dt in doctypes:
+        # We only validate if both product and docset are explicitly provided and are not wildcards
+        if dt.product and "*" not in dt.product and dt.docset and "*" not in dt.docset:
+            prod_val = dt.product.value
+            # Fetch all valid 'path' attributes from docsets under this specific product
+            valid_docsets = tree.xpath(f"//product[@id='{prod_val}']/docset/@path")
+
+            # Format them for display, always prepending the wildcard allowed value
+            valid_docsets_str = ["*", *sorted(str(v) for v in valid_docsets)]
+
+            for ds in dt.docset:
+                if ds not in valid_docsets:
+                    allowed_str = ", ".join(f"'{v}'" for v in valid_docsets_str)
+                    console.print("[red]Error parsing doctype: 1 validation error for Doctype[/red]")
+                    console.print("docset")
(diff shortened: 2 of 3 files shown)


sushant-suse:
Fixed all things that I think noteworthy from Copilot.
Waiting for the final review from @tomschr before merging it.

changelog.d/303.bugfix.rst
```diff
@@ -0,0 +1 @@
+Added dynamic validation to the ``portal list`` command to provide helpful error messages with available allowed values when an invalid docset is provided.
```
tomschr:
```suggestion
Added dynamic validation to the :command:`portal list` command to provide helpful error messages with available allowed values when an invalid docset is provided.
```

tests/cli/cmd_portal/test_cmd_list.py
```diff
@@ -46,6 +46,45 @@ def test_portal_list_invalid_doctype() -> None:
     assert "Error parsing doctype:" in result.output
 
 
+@patch("docbuild.cli.cmd_portal.cmd_list.parse_portal_config", new_callable=AsyncMock)
```
tomschr: Prefer `@patch.object` over `@path` here. The regular `@patch` decorator requires a string and that is problematic when refactoring with VSCode.

According to my experiences when refactoring with VSCode, having a `@patch.object` makes it more likely that the editor gets it right.

src/docbuild/cli/cmd_portal/cmd_list.py
```diff
@@ -172,6 +172,29 @@ def print_hierarchy(
...
+def validate_docsets_against_xml(
+    doctypes: list[Doctype], tree: etree._ElementTree | etree._Element, console: Console
+) -> None:
+    """Dynamically validate that provided docsets exist for their respective products."""
+    for dt in doctypes:
+        # We only validate if both product and docset are explicitly provided and are not wildcards
+        if dt.product and "*" not in dt.product and dt.docset and "*" not in dt.docset:
+            prod_val = dt.product.value
+            # Fetch all valid 'path' attributes from docsets under this specific product
+            valid_docsets = tree.xpath(f"//product[@id={prod_val!r}]/docset/@path")
```
tomschr: Technically it's correct. However, I would try to avoid using XPath expression directly on the tree. If we change the XML config, we would have to deal with all these spread out XPath expressions.

For that reason, I'd suggest to use `Deliverable`. It's made for these things.😉 You could have a look in `src/docbuild/cli/cmd_portal/cmd_list.py:230`. 

In the header, add the line `from docbuild.config.xml.list import list_all_deliverables`, then:


```suggestion
    for node in list_all_deliverables(tree, doctypes):
       deli = Deliverable(_node=node)
       # ... Check the 
       # Get the docset with deli.xml.docsetid
```

To be honest, I'm not completely sure if this is feasible.

src/docbuild/cli/cmd_portal/cmd_list.py
```diff
@@ -172,6 +172,29 @@ def print_hierarchy(
...
+            # Format them for display, always prepending the wildcard allowed value
+            valid_docsets_str = ["*", *sorted(str(v) for v in valid_docsets)]
+
+            for ds in dt.docset:
+                if ds not in valid_docsets:
+                    allowed_str = ", ".join(f"'{v}'" for v in valid_docsets_str)
+                    console.print("[red]Error parsing doctype:[/red] 1 validation error for Doctype")
+                    console.print("docset")
+                    console.print(f"  Value error, '{ds}' is not a valid Docset. Allowed values are: {allowed_str}")
+                    raise click.Abort()
```
tomschr: That is correct, but may create a small usability issue.

Imagine someone calls the tool like this:

```
docbuild --env-config ../main/env.devel.toml portal list 'suma/x' 'suse-ai/y'
```

Both doctypes are invalid as there is neither a `suma/x` nor a `suse-ai/y` docset. But if you run it, you get:

```
Error parsing doctype: 1 validation error for Doctype
docset
  Value error, 'x' is not a valid Docset. Allowed values are: '*', '4.3', '5.0', '5.1', '5.2'
Aborted!
```

The function should list _all_ invalid docsets, not just the first one. Maybe something like this:

```
Error parsing doctype: 2 validation errors for Doctype:

* suma/x is not a valid Docset.
  Allowed values are: '*', '4.3', '5.0', '5.1', '5.2'
* suse-ai/y is not a valid Docset.
  Allowed values are: '*', '1.0'

Aborted!
```

tomschr (changes requested):
Thanks for your contribution! 👍

I have some ideas regarding the validation and a usability issue. Let me know if you have questions. 🙂

tests/cli/cmd_portal/test_cmd_list.py
sushant-suse: Thanks for details Toms.
Done, updated `patch` with `patch.object`

src/docbuild/cli/cmd_portal/cmd_list.py
sushant-suse: You were spot on about the architectural boundary. Sprinkling raw XPaths into the CLI logic is a bad habit and makes schema changes a nightmare.

I successfully implemented your `Deliverable` approach, but I had to navigate two small hurdles (which explains your hesitation about whether it was feasible!):

1. **Pydantic Strict Typing:** I couldn't manually instantiate the `Doctype` with raw wildcard strings (`["*"]`) because Pydantic strictly enforces internal `LanguageCode` and `LifecycleFlag` types. I bypassed this by using the class's own string parser: `Doctype.from_str(f"{prod_val}/*/*")`, which handles all the deep typing for us!
2. **Empty Docsets:** Because `list_all_deliverables` yields `<deliverable>` nodes, it naturally filters out any docsets that are completely empty. This is actually a *better* user experience (why suggest an empty docset as an allowed option?), but it meant I had to update our test XML to ensure the dummy `15sp5` docset actually contained a deliverable so the assertions would pass.

src/docbuild/cli/cmd_portal/cmd_list.py
sushant-suse: Excellent catch on the UX Toms. Failing fast on the first error is definitely annoying when you are passing multiple arguments.

I've updated the validation loop to act as an error collector. It now processes all the provided doctypes, aggregates any invalid docsets, and prints a comprehensive, bulleted report exactly like your suggested output before aborting. I also updated the test assertions to match the new consolidated format.

Thanks for pushing the UX on this one, it looks much more professional now.

tomschr (approved):
Thank you Sushant! We can merge it. 👍 
Ship it!

PR #349, opened 2026-06-18
docs: differentiate shell commands from their output using '$'

yacar-suse:
in order to mitigate potential reader confusion.
This has already been done in other doc files e.g. `user/portal-config/validate-portal.rst`.
These changes unify the practice across all shell code blocks.


sushant-suse (approved):
Good catch @yacar-suse. Thanks.

tomschr (approved):
Great, thank to all of you! :+1:

PR #350, opened 2026-06-23
docs: create common set of AI agent skills

### LINKED ISSUE #266: Create a common set of prompts, skill sets
### Situation

We have `.github/agents/docbuild.agent.md` for working with AI bots, but maybe we lack some skill files.

### Use Case

Having a common ground.

#### Prompts
* Create similar prompts under `.github/agents/` for XSLT, RNG, and documentation. Or are they unnecessary due to skill files?

#### Skill sets
 
* Create skill sets.

  As different folders exist where AI bots look at skills (`.github/skills/`, `.agents/skills/`, `.claude/skills/`), the most(?) general one seems to be `.agents/skills/`.
  We create a directory `.agents/skills/` and link from `.github/skills/` to it.
* Create a new directory that describes the skill under `.agents/skills/`.
* Create a file `SKILL.md` and describe the skill. We should create a skill for:
  * how to run the specific task with the `docbuild` tool via the custom alias (`devel/activate-aliases.sh`)
  * how to run the test suite via `upytest` and how to interpret the output.
  * how to create news fragments via `towncrier`.
  * how to create the documentation, where to find the sources, and where to look at the output.
  * Something else?
* Mention it in our documentation (Developer Guide)?

It's recommended to follow these points:

* Keep it short (under 500 lines¹).
* Use coherent units (tackle only one skill).
* Provide defaults, not too much options.
* Favor procedures.
* If feasible, provide a checklist.
* Instruct the agent to validate its own work before moving on.

### Additional information

1. https://agentskills.io/
2. https://github.com/agentskills/agentskills
3. https://github.com/VoltAgent/awesome-agent-skills
4. https://github.com/forrestchang/andrej-karpathy-skills/blob/main/skills/karpathy-guidelines/SKILL.md

sushant-suse:
Closes #266

Created a standardized set of AI agent "skill" files to help AI assistants interact with the repository's workflows more effectively and reliably.

### Changes Made

* Created `.agents/skills/` directory with detailed `SKILL.md` guides for:
  * **Towncrier**: Managing news fragments.
  * **Testing**: Running tests via `upytest` and analyzing failures.
  * **Docbuild Alias**: Activating and using the `devel/activate-aliases.sh` script.
  * **Documentation**: Building Sphinx docs via `makedocs`.
* Added a symlink at `.github/skills` pointing to `.agents/skills/` for broader tool compatibility.
* Updated `.github/agents/docbuild.agent.md` to instruct the AI to consult these skill files.
* Added news fragment `266.feature.rst`.

### Self-Review Checklist

- [x] Skill files are concise, procedural, and include checklists/validation steps.
- [x] Symlink created successfully.

**Documentation & CI**:

### DIFF
--- a/.agents/skills/docbuild-alias/SKILL.md
+++ b/.agents/skills/docbuild-alias/SKILL.md
@@ -1,0 +1,21 @@
+# Skill - Running the docbuild tool via Custom Aliases
+
+## Context
+
+This repository provides custom shell aliases to make running tasks easier without needing to type long `uv run` commands constantly. 
+
+## Procedure
+
+1. Before attempting to run custom repository commands, you must source the alias script.
+2. Run: `source devel/activate-aliases.sh`
+3. Once sourced, you can use the custom aliases available in the environment (for example, `docbuild`, `upytest`, `makedocs`).
+4. To run the main CLI tool, simply use the `docbuild` alias followed by the relevant command (for example, `docbuild portal list`).
+
+## Checklist
+
+- [ ] Did you run `source devel/activate-aliases.sh` in the current terminal session?
+- [ ] Are you using the alias directly instead of invoking python/uv manually?
+
+## Validation
+
+Run `alias` or `type docbuild` in the terminal to verify the alias is active before attempting to build or test.
--- a/.agents/skills/docbuild-alias/SKILL.md
+++ b/.agents/skills/docbuild-alias/SKILL.md
@@ -9,8 +9,8 @@ This repository provides custom shell aliases to make running tasks easier witho
 1. Before attempting to run custom repository commands, you must source the alias script.
 2. Run: `source devel/activate-aliases.sh`
 3. Once sourced, you can use the custom aliases available in the environment (for example, `docbuild`, `upytest`, `makedocs`).
-4. **CRITICAL**: To run the main CLI tool, you must *always* use a dedicated development environment configuration to avoid conflicting with or polluting the developer's local host state. 
-5. Always run commands using the `--env-config` flag pointing to the development config (for example, `docbuild --env-config env.development.toml portal list`). Do not run `docbuild` without an isolated environment configuration.
+4. **CRITICAL**: To run the main CLI tool, you must *always* use the dedicated agent environment configuration to avoid conflicting with or polluting the human developer's local host state.
+5. Always run commands using the `--env-config` flag pointing to the agent config: `docbuild --env-config env.agent.toml <command>`. Do not run `docbuild` without this isolated environment configuration.
 
 ## Checklist
 
--- a/.agents/skills/docbuild-alias/SKILL.md
+++ b/.agents/skills/docbuild-alias/SKILL.md
@@ -9,7 +9,8 @@ This repository provides custom shell aliases to make running tasks easier witho
 1. Before attempting to run custom repository commands, you must source the alias script.
 2. Run: `source devel/activate-aliases.sh`
 3. Once sourced, you can use the custom aliases available in the environment (for example, `docbuild`, `upytest`, `makedocs`).
-4. To run the main CLI tool, simply use the `docbuild` alias followed by the relevant command (for example, `docbuild portal list`).
+4. **CRITICAL**: To run the main CLI tool, you must *always* use a dedicated development environment configuration to avoid conflicting with or polluting the developer's local host state. 
+5. Always run commands using the `--env-config` flag pointing to the development config (for example, `docbuild --env-config env.development.toml portal list`). Do not run `docbuild` without an isolated environment configuration.
 
 ## Checklist
 
--- a/.agents/skills/documentation/SKILL.md
+++ b/.agents/skills/documentation/SKILL.md
@@ -1,4 +1,4 @@
-# Skill -  Creating and Building Documentation
+# Skill - Creating and Building Documentation
 
 ## Context
 
--- a/.agents/skills/documentation/SKILL.md
+++ b/.agents/skills/documentation/SKILL.md
@@ -1,0 +1,25 @@
+# Skill -  Creating and Building Documentation
+
+## Context
+
+This project uses Sphinx for documentation. The source files are written in reStructuredText (.rst) and the configuration is handled via `conf.py`.
+
+## Procedure
+
+1. Locate the documentation source files in `docs/source/`.
+2. Edit or add `.rst` files as needed to document new features or CLI commands.
+3. If adding a new file, ensure it is referenced in the `toctree` of an existing index file.
+4. Ensure the custom aliases are active (`source devel/activate-aliases.sh`).
+5. Build the documentation by running the custom alias: `makedocs`.
+6. The HTML output is generated and placed in the appropriate build output directory (usually `docs/build/html/`).
+
+## Checklist
+
+- [ ] Are changes placed inside `docs/source/`?
+- [ ] Did you use standard Sphinx-style reStructuredText?
+- [ ] Did you build the docs using the `makedocs` alias?
+- [ ] Did the build complete without Sphinx warnings?
+
+## Validation
+
+Always run `makedocs` after modifying `.rst` files or docstrings. Do not consider a documentation task complete if the Sphinx build emits warnings or errors.
--- a/.agents/skills/documentation/SKILL.md
+++ b/.agents/skills/documentation/SKILL.md
@@ -2,7 +2,9 @@
 
 ## Context
 
-This project uses Sphinx for documentation. The source files are written in reStructuredText (.rst) and the configuration is handled via `conf.py`.
+This project uses Sphinx for documentation. The source files are written in reStructuredText (.rst) and the configuration is handled via `conf.py`. 
+
+**Note on Structure**: The documentation is separated into different guides (for example, user, developer, etc.). When reading or updating documentation, always ensure you are navigating and modifying the correct guide for the target audience.
 
 ## Procedure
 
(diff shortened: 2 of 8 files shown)


sushant-suse:
Fixed all things that I think noteworthy from Copilot.
Waiting for the final review from @tomschr before merging it.

.agents/skills/docbuild-alias/SKILL.md
```diff
@@ -0,0 +1,21 @@
...
+## Context
+
+This repository provides custom shell aliases to make running tasks easier without needing to type long `uv run` commands constantly. 
+
+## Procedure
+
+1. Before attempting to run custom repository commands, you must source the alias script.
+2. Run: `source devel/activate-aliases.sh`
+3. Once sourced, you can use the custom aliases available in the environment (for example, `docbuild`, `upytest`, `makedocs`).
+4. To run the main CLI tool, simply use the `docbuild` alias followed by the relevant command (for example, `docbuild portal list`).
```
tomschr: That is technically correct. However, that would use the hard-coded values for the environment.

Should we provide a env config solely for an agent? 🤔 I'm not entirely sure. Just a crazy idea we can discuss.
My concern is, that if we let an agent run `docbuild` _without_ a dedicated env config, it may conflict with existing paths in env config files.

.agents/skills/documentation/SKILL.md
```diff
@@ -0,0 +1,25 @@
+# Skill - Creating and Building Documentation
+
+## Context
+
+This project uses Sphinx for documentation. The source files are written in reStructuredText (.rst) and the configuration is handled via `conf.py`.
```
tomschr: Not sure if it's needed or useful, but should we tell that the documentation is separated between different guides?

.agents/skills/testing/SKILL.md
```diff
@@ -0,0 +1,22 @@
...
+## Context
+
+This repository uses `pytest` for testing, wrapped in a custom alias script to ensure the environment is correctly configured before tests run. 
+
+## Procedure
+
+1. Before running tests, ensure the development aliases are active in your session.
+2. Run the test suite using the project-specific alias `upytest` (instead of standard `pytest`).
+3. If running a specific test file, append the path: `upytest tests/path/to/test.py`.
+4. Analyze the output. If tests fail, read the traceback carefully, specifically looking for assertion errors, missing mocks, or formatting mismatches (like unexpected newline characters).
```
tomschr: Perhaps we should tell the AI to use the `-v` option to get more elaborate output?

.agents/skills/testing/SKILL.md
```diff
@@ -0,0 +1,22 @@
...
+
+## Checklist
+
+- [ ] Did you use `upytest` instead of standard `pytest`?
+- [ ] If tests failed, did you trace the failure back to the exact line in the test file or source code?
+- [ ] Did you strip or handle terminal formatting or newlines in string assertions if testing rich CLI output?
+
```
tomschr, line 20: Maybe add another line regarding coverage reports.

The AI must _ignore_ coverage reports when using a specific test file. Calling `upytest TEST_FILE(S)` _fails_ due to a coverage lower than 90%. This is normal and not an error.

.agents/skills/towncrier/SKILL.md
```diff
@@ -0,0 +1,24 @@
+# Skill - Managing News Fragments (Towncrier)
+
+## Context
+
+This repository uses `towncrier` to generate the changelog. Every pull request that introduces a user-facing change, bugfix or significant internal change must include a news fragment.
+
+## Procedure
+
+1. Identify the issue number associated with your current task.
+2. Determine the correct fragment type. Allowed types are: `breaking`, `bugfix`, `deprecation`, `doc`, `feature`, `refactor`, `removal`, `infra`, `security`.
```
tomschr: That's all correct. Hhowever, I'm only concerned that the list will deviate from the `towncrier.toml` file at some point.

Not sure which is the best practice here, but either let read the AI the `towncrier.toml` file. This is the ultimate source of all available fragment types. Or we leave it as is and add a comment in the config file: when something changes, also adjust the skill file.

tomschr (changes requested):
Great work, Sushant! 👍

I have some ideas. To be honest, I'm not completely sure if this is "too much" for the AI or it's a good addition.

.agents/skills/docbuild-alias/SKILL.md
sushant-suse: That isn't a crazy idea at all, it is actually a crucial safety guardrail. You are completely right. I updated the `SKILL.md` file to add a "CRITICAL" instruction. It now explicitly tells the agent that it must *always* append the `--env-config` flag pointing to a development configuration (like `env.devel.toml`) to sandbox itself and avoid polluting the host state.

Good catch on the environmental isolation. Thanks.

.agents/skills/testing/SKILL.md
sushant-suse: That is a really smart addition. I've added a new "CRITICAL" item to the checklist explicitly instructing the AI to ignore coverage threshold failures when running targeted test files. This will save it from a lot of unnecessary hallucinations.

.agents/skills/documentation/SKILL.md
sushant-suse: Updated thanks.

.agents/skills/testing/SKILL.md
sushant-suse: Absolutely. I've updated Step 3 to explicitly instruct the agent to append `-v` when targeting a test file.

.agents/skills/towncrier/SKILL.md
sushant-suse: Between the two options, instructing the AI to read the `towncrier.toml` file is definitely the best practice. I've removed the hardcoded list and updated Step 2 to explicitly instruct the agent to read `towncrier.toml` to determine the allowed fragment types. Great catch for long-term maintainability. Thanks Toms.

.agents/skills/docbuild-alias/SKILL.md
```diff
@@ -0,0 +1,22 @@
...
+
+This repository provides custom shell aliases to make running tasks easier without needing to type long `uv run` commands constantly. 
+
+## Procedure
+
+1. Before attempting to run custom repository commands, you must source the alias script.
+2. Run: `source devel/activate-aliases.sh`
+3. Once sourced, you can use the custom aliases available in the environment (for example, `docbuild`, `upytest`, `makedocs`).
+4. **CRITICAL**: To run the main CLI tool, you must *always* use a dedicated development environment configuration to avoid conflicting with or polluting the developer's local host state. 
+5. Always run commands using the `--env-config` flag pointing to the development config (for example, `docbuild --env-config env.development.toml portal list`). Do not run `docbuild` without an isolated environment configuration.
```
tomschr: Should we more explicit here? I was thinking about creating a dedicated `env.agent.toml` file in the project's root so it can be used here?

.agents/skills/documentation/SKILL.md
```diff
@@ -0,0 +1,27 @@
...
+4. Ensure the custom aliases are active (`source devel/activate-aliases.sh`).
+5. Build the documentation by running the custom alias: `makedocs`.
+6. The HTML output is generated and placed in the appropriate build output directory (usually `docs/build/html/`).
+
+## Checklist
+
+- [ ] Are changes placed inside `docs/source/`?
+- [ ] Did you use standard Sphinx-style reStructuredText?
+- [ ] Did you build the docs using the `makedocs` alias?
+- [ ] Did the build complete without Sphinx warnings?
```
tomschr: That _might_ be dangerous as we currently have warnings in our docs. Maybe we should replace warnings with errors?

tomschr (changes requested):
I have only two ideas/suggestions, otherwise it's great! 👍

.agents/skills/docbuild-alias/SKILL.md
sushant-suse: I love that idea. I have updated the `SKILL.md` to explicitly enforce `docbuild --env-config env.agent.toml <command>` and added a base `env.agent.toml` file (which comments) to the repository root so the agent has its own dedicated sandbox ready to go.

.agents/skills/documentation/SKILL.md
sushant-suse: That is a great point Toms. I have updated the checklist to focus strictly on **errors** and added a small note explicitly telling the agent that it is safe to ignore pre-existing warnings. This should keep it focused strictly on its actual task.

env.agent.toml
```diff
@@ -0,0 +1,2 @@
+# Dedicated environment configuration for AI agents.
+# This file ensures agent actions do not pollute local developer environments.
```
tomschr, line 2: Good start! 👍 My idea was that we actually fill the env config with some keys like:

* `server.name = "ai-agent"`
  We want to give it a dedicated name so we can identify it.
* `server.role = "devel"`
  Certainly not production. 😉 
* `root_config_dir = "~/.config/docbuild/"`
  The home directory does not have any restrictions on permission.
* `paths.base_cache_dir = "~/.local/state/docbuild"`
  Same for the cache

I think that should be the minimum keys we need. As we merged the issue with the partial config, it should work. 😊

tomschr (commented):
Thanks! I left a small comment. 😺

env.agent.toml
sushant-suse: Ah, that makes perfect sense. I love the `server.name = "ai-agent"` flag so we can immediately identify agent-generated logs. I have populated `env.agent.toml` with those exact keys and pushed the update.

tomschr (approved):
As always, great work Sushant! 🎉 Let's merge it. 👍

PR #354, opened 2026-07-06
docs: describe uv installation on (open)suse and link to uv docs

yacar-suse:
for other operating systems, as likely most users of docbuild will use a suse distro and recommending a 'curl | sh' installer is bad practice.

### DIFF
--- a/docs/source/user/install.rst
+++ b/docs/source/user/install.rst
@@ -14,67 +14,61 @@ To install the docbuild tool, follow these steps:
 
 .. _prepare-installation:
 
-Preparing for Installation
+Preparing environment
 --------------------------
 
-It is highly recommended to use the `package and project manager uv <https://docs.astral.sh/uv>`_ to install the docbuild tool. This ensures that all dependencies are managed correctly and that the installation is reproducible across different environments.
-
-There are different methods to install the :command:`uv` package manager
-(see `Installing uv <https://docs.astral.sh/uv/getting-started/installation/>`_). In this case, use the standalone installer:
-
+It is highly recommended to use the `package and project manager uv <https://docs.astral.sh/uv>`_
+to install the docbuild tool. This ensures that all dependencies are managed correctly and that
+the installation is reproducible across different environments.
 
 1. **Install uv**
 
-   Run the following command in your terminal:
+   There are different methods to install :command:`uv`. Preferably use your systems package manager,
+   otherwise follow the `uv documentation <https://docs.astral.sh/uv/getting-started/installation/>`_.
+
+   For (open)SUSE Linux run the following command in your terminal:
 
    .. code-block:: shell-session
 
-      $ curl -sSfL https://astral.sh/uv/install.sh | sh
-
-
-   This will install two commands :command:`uv` and :command:`uvx` in :file:`~/.local/bin/`. Make sure this directory is in your :envvar:`PATH` environment variable.
+      $ sudo zypper install python313-uv
 
 2. **Check the installation**
 
-   After the installation, verify that :command:`uv` is installed correctly by running:
+   After the installation, verify that :command:`uv` is installed and in your PATH by running:
 
    .. code-block:: shell-session
 
       $ type uv
-      uv is hashed (/home/tux/.local/bin/uv)
+      uv is hashed (/path/to/uv)
       $ uv --version
       ...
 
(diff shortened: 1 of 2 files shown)


docs/source/user/install.rst
```diff
@@ -17,60 +17,54 @@ To install the docbuild tool, follow these steps:
...
 2. **Check the installation**
 
-   After the installation, verify that :command:`uv` is installed correctly by running:
+   After the installation, verify that :command:`uv` is installed and in your PATH by running:
 
    .. code-block:: shell-session
 
       $ type uv
-      uv is hashed (/home/tux/.local/bin/uv)
+      uv is hashed (/path/to/uv)
```
tomschr, line 42: If you install it with a package manager, it will probably be available there:

```suggestion
      uv is hashed (/usr/bin/uv)
```

docs/source/user/install.rst
```diff
@@ -17,60 +17,54 @@ To install the docbuild tool, follow these steps:
 Preparing for Installation
```
tomschr: Hmn... maybe we should be more precise? Something like this:


```suggestion
Installing the uv Package Manager
```

What do you think?

docs/source/user/install.rst
tomschr: Was not your fault, but I think, we should be more precise here too:

```suggestion
Installing docbuild
```

tomschr: I forgot the correct Sphinx macro:


```suggestion
   This will create a virtual environment in the directory :file:`.venv`.
```

tomschr: I would remove the "...using pip" as it's actually `uv sync`:

```suggestion
   Ensure you have Python 3.12 or higher installed, then install the required dependencies:
```

docs/source/user/install.rst
```diff
@@ -108,7 +103,11 @@ Installing the tool
       Built docbuild @ file:///.../docbuild
       Installed 15 packages in 2.11s
 
-   The `--frozen` flag is used here to ensure that `uv` installs dependencies exactly as specified in the `uv.lock` file. This is crucial for **reproducible builds** and maintaining a **consistent environment** across different machines or deployments, as it prevents `uv` from attempting to resolve and potentially update dependency versions.
+   The `--frozen` flag is used here to ensure that `uv` installs dependencies
```
tomschr: RST/Sphinx requires two backticks; also adding the `:command:` macro:

```suggestion
   The ``--frozen`` flag is used here to ensure that :command:`uv` installs dependencies
```

docs/source/user/install.rst
```diff
@@ -108,7 +103,11 @@ Installing the tool
...
+   exactly as specified in the `uv.lock` file. This is crucial for
```
tomschr: Add a `:file:` macro:

```suggestion
   exactly as specified in the :file:`uv.lock` file. This is crucial for
```

docs/source/user/install.rst
```diff
@@ -108,7 +103,11 @@ Installing the tool
...
+   **reproducible builds** and maintaining a **consistent environment** across
+   different machines or deployments, as it prevents `uv` from attempting to
```
tomschr: Same here:

```suggestion
   different machines or deployments, as it prevents :command:`uv` from attempting to
```

tomschr (changes requested):
Thank you @yacar-suse, great work! That's a valid point that I haven't thought of. 👍

I've added comments, but they are minor things only.

docs/source/user/install.rst
yacar-suse: I tought it would make sense to generalize here, in case someone is using the uv installer instead.

docs/source/user/install.rst
yacar-suse: Maybe like this:
```suggestion
Preparing environment
```

docs/source/user/install.rst
tomschr: Yes, that's actually a good point considering MacOS or other means. Then just leave it as is. :+1:

docs/source/user/install.rst
tomschr: Sounds good, use it! :+1:

tomschr:
@yacar-suse You can (squash) merge this PR. Have fun clicking the button! :smile:

PR #367, opened 2026-07-10
ci: fix build-and-deploy failure

yacar-suse:
towncrier failed because the options '--yes' and '--keep' were selected at the same time.


tomschr:
Ahh, good catch! Thank you Yunus! :+1:

tomschr (approved):

PR #369, opened 2026-07-14
Reorg XDG directory setup and defaults

tomschr:
Reason for this PR was to have a config that its paths are based on the user's home directory. It also fixes some oddities with log and runtime directories (mainly lock files).

* Reorg in `constants.py`:
  * Use module `platformdirs` to use a consistent
  * Fix some `tuple[str, ...]` types
  * Update `VALID_PRODUCTS`
  * Introduce XDG directory setup variables for state, config, data, cache, and runtime.
  * Derive `BASE_STATE_DIR` from `STATE_HOME` of platformdirs, not from `Path.home()`.
  * Use `RUNTIME_DIR / "locks"` for locking files

* Reorg in `defaults.py`:
  * Adjust `DEFAULT_ENV_CONFIG` to use XDG variables from `constants.py` or use config placeholders to allow customization in user or repo configs
  * Use home directory as a place for storage as this is writeable and doesn't need additional write privileges.
  * Fix logging directory in `paths.tmp.log_dir` key. Previously, it was in `STATE_HOME/log`, but it makes more sense to have a `server.name` part in between to distinguish between different names.

* Additions in `env.py`:
  * Add `json_cache_dir`: Cache specifically for JSON data used in the portal.
  * Add `runtime_base_dir`: Base Runtime Directory (Per-Run) We didn't have that before
  * Add `lock_dir`: Directory for lock files used to prevent concurrent builds, usually for lock files. Is
  * Disable `base_tmp_dir` as it's not really needed; we have already `paths.tmp.tmp_base_dir`.

* Add missing keys in `env.example.toml`; make order follow the file `defaults.py`

* Adjust test cases

### DIFF
--- a/changelog.d/369.refactor.rst
+++ b/changelog.d/369.refactor.rst
@@ -1,0 +1,1 @@
+The project's directory structure was refactored to align with XDG standards using ``platformdirs``, centralizing path definitions and introducing new, more flexible configuration options for cache, runtime, and lock directories.
--- a/docs/source/reference/_autoapi/docbuild/cli/cmd_portal/cmd_list/index.rst
+++ b/docs/source/reference/_autoapi/docbuild/cli/cmd_portal/cmd_list/index.rst
@@ -15,6 +15,14 @@ Functions
 .. autoapisummary::
 
    docbuild.cli.cmd_portal.cmd_list.build_hierarchy
+   docbuild.cli.cmd_portal.cmd_list.parse_doctypes
+   docbuild.cli.cmd_portal.cmd_list.get_display_name
+   docbuild.cli.cmd_portal.cmd_list.append_translations
+   docbuild.cli.cmd_portal.cmd_list.append_formats
+   docbuild.cli.cmd_portal.cmd_list.append_categories
+   docbuild.cli.cmd_portal.cmd_list.append_repo
+   docbuild.cli.cmd_portal.cmd_list.build_deliverable_branch
+   docbuild.cli.cmd_portal.cmd_list.print_hierarchy
    docbuild.cli.cmd_portal.cmd_list.async_list_cmd
    docbuild.cli.cmd_portal.cmd_list.list_cmd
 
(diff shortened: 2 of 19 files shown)


tomschr:
@copilot Check again.

sushant-suse:
Excellent architectural refactor, Toms. Moving to `platformdirs` and adopting XDG Base Directory standards is a massive improvement. Shifting the ephemeral lock files to the runtime directory (`/run/user/...`) is a particularly brilliant safety measure to prevent stale locks from persisting across reboots.

The string interpolation updates in the defaults engine are also much safer now.

I just found one minor bug/issue/question regarding dynamic resolution, and one tiny syntax nitpick. Once those are tweaked, this is absolutely good to go!

**1. `log_dir` won't dynamically update (defaults.py)**:
In `src/docbuild/cli/defaults.py`, you mapped the log directory like this:
`"log_dir": BASE_LOG_DIR,`

Because `BASE_LOG_DIR` is strictly evaluated in `constants.py` using `DEFAULT_SERVER_NAME`, it won't dynamically adjust if a user overrides `server.name` in their TOML file. (Contrast this with `backup_dir` on line 85, which correctly uses `f"{STATE_HOME}/{{server.name}}/backup"`).
We should probably update `log_dir` in `defaults.py` to use the `{server.name}` placeholder so it properly isolates logs per environment!

**2. Path composition nitpick (constants.py)**:
Around line 176 in `constants.py`, `BASE_LOG_DIR: Path = Path(f"{STATE_HOME}/{DEFAULT_SERVER_NAME}/log")`
Right below it, `BASE_STATE_DIR` correctly uses the standard Pathlib `/` operator. We should probably avoid string interpolation (`f"{}"`) when combining `Path` objects:
`BASE_LOG_DIR: Path = STATE_HOME / DEFAULT_SERVER_NAME / "log"`

Tests and models look completely solid. Great work on this.

tomschr:
> 1. log_dir won't dynamically update (defaults.py):
> [...]
> it won't dynamically adjust if a user overrides `server.name` in their TOML file.

You hit the nail on the top! :+1: I've tried to solve this by going back to the original line:

```python
"log_dir": f"{STATE_HOME}/{{server.name}}/log",
```

The server name will evaluate lazily, but `STATE_HOME` will be replaced.

I've tried it and it seems to work:

* Without any env config:

  ```shell
  $ docbuild config list | grep 'log_dir'
      "log_dir": "/home/toms/.local/state/docbuild/default-env/log"
  ```

* With an env config:

  ```toml
  # env.test.toml
  [server]
  name = "test"
  ```

  ```shell
  $ docbuild --env-config env.test.toml config list | grep 'log_dir'
      "log_dir": "/home/toms/.local/state/docbuild/test/log"
  ```

Would that work?

sushant-suse:
> Excellent architectural refactor, Toms. Moving to `platformdirs` and adopting XDG Base Directory standards is a massive improvement. Shifting the ephemeral lock files to the runtime directory (`/run/user/...`) is a particularly brilliant safety measure to prevent stale locks from persisting across reboots.
> 
> The string interpolation updates in the defaults engine are also much safer now.
> 
> I just found one minor bug/issue/question regarding dynamic resolution, and one tiny syntax nitpick. Once those are tweaked, this is absolutely good to go!
> 
> **1. `log_dir` won't dynamically update (defaults.py)**: In `src/docbuild/cli/defaults.py`, you mapped the log directory like this: `"log_dir": BASE_LOG_DIR,`
> 
> Because `BASE_LOG_DIR` is strictly evaluated in `constants.py` using `DEFAULT_SERVER_NAME`, it won't dynamically adjust if a user overrides `server.name` in their TOML file. (Contrast this with `backup_dir` on line 85, which correctly uses `f"{STATE_HOME}/{{server.name}}/backup"`). We should probably update `log_dir` in `defaults.py` to use the `{server.name}` placeholder so it properly isolates logs per environment!
> 
> **2. Path composition nitpick (constants.py)**: Around line 176 in `constants.py`, `BASE_LOG_DIR: Path = Path(f"{STATE_HOME}/{DEFAULT_SERVER_NAME}/log")` Right below it, `BASE_STATE_DIR` correctly uses the standard Pathlib `/` operator. We should probably avoid string interpolation (`f"{}"`) when combining `Path` objects: `BASE_LOG_DIR: Path = STATE_HOME / DEFAULT_SERVER_NAME / "log"`
> 
> Tests and models look completely solid. Great work on this.

Yes, that is absolutely perfect!

Using `{{server.name}}` inside the f-string correctly escapes the braces so the config engine can lazy-evaluate it later. Your test output proves it works exactly as intended and perfectly isolates the logs per environment.

tomschr:
@sushant-suse I changed a bit of the logging. Before this change it was this:

1. Load the app config
2. Call `setup_logging`
3. Load the env config

I had to change this order to allow defining a log path in the env config and set up the logging _after_ the env config was loaded. Here are the summary:

1. In `cmd_cli.py` I changed the order. First we load app and env configs. Then I pass the `log_dir` (which is located in the env config) to the `setup_logging`.
1. The `setup_logging()` in `logging.py` accepts a `log_dir` argument now. This was needed to dynamically add the logging directory.
1. I removed the now obosolete `BASE_LOG_DIR`. Same with `create_base_log_dir` as this is created by Pydantic's model. Adjusted the tests accordingly.

sushant-suse:
> @sushant-suse I changed a bit of the logging. Before this change it was this:
> 
> 1. Load the app config
> 2. Call `setup_logging`
> 3. Load the env config
> 
> I had to change this order to allow defining a log path in the env config and set up the logging _after_ the env config was loaded. Here are the summary:
> 
> 1. In `cmd_cli.py` I changed the order. First we load app and env configs. Then I pass the `log_dir` (which is located in the env config) to the `setup_logging`.
> 2. The `setup_logging()` in `logging.py` accepts a `log_dir` argument now. This was needed to dynamically add the logging directory.
> 3. I removed the now obosolete `BASE_LOG_DIR`. Same with `create_base_log_dir` as this is created by Pydantic's model. Adjusted the tests accordingly.

---

Solving the problem by loading the environment config *before* initializing the logger is exactly the right move. I also love that you ripped out `BASE_LOG_DIR` and `create_base_log_dir`—relying on Pydantic's `EnsureWritableDirectory` to automatically create the folder during validation is a much cleaner, framework-native approach.

Injecting the `log_dir` directly into `setup_logging()` also makes that function way easier to test in the future. This looks completely solid to me. Fantastic work on this entire PR Toms.

PR #383, opened 2026-07-21
feat: Add runtime system dependency validation and doctor CLI command

### LINKED ISSUE #234: Investigate how to express system dependencies
### Situation

The docbuild tool needs not only Python libraries, but also other tools like `daps`, `jing`, or `trang`.
Additionally, there are second order dependencies from the primary ones like Java, `xsltproc` or `xmllint`.

### Use Case

First, we need to define the scope:

* **First-order dependencies**

  These are the tools that are needed _directly_ by docbuild. For example, we need `daps` to build our deliverables.
  A tool/script can be found via `shutils.which`.

* **Second-order dependencies**

  These are the tools that depend on the first-order dependencies. For example, `daps` uses `xmllint` and `xsltproc` to process the XML code. Sometimes these kind of dependencies are more "abstract", like Java, GlibC etc. These makes it harder to find them as it's dependent on the system (Linux vs. MacOS) and the distro (openSUSE/RPM-based vs. Debian/DEB based).

This issue is about first-order dependencies. 

We will _never_ be able to determine with absolute certainty whether all dependencies are met. That's not the goal. But at least to find some problems before the script is doing it's work may be helpful.

The scope is to:

* First check all first-order dependencies and give meaningful error messages when it's not met.
* If a second-order dependency isn't found, raise an exception. In the error message, try to be as precise as possible.

### Possible Implementation

We could use a custom build backend based on PEP 517.

1. Change the `pyproject.toml` file:

   1. Add a `tool.docbuild.system-requiremnt` key that holds the tool and an optional version:

       ```toml
       [tool.docbuild]
       system-requirements = [
          "jing >= 20220510",
          "trang",            # Any version is fine
          "daps >= 4",
          "xmllint",          # Any version is fine
          "xsltproc"
       ]
       ```

    2. Adjust the `[build-system]` section:

       ```toml
       [build-system]
       requires = ["setuptools", "wheel",]
       build-backend = "_docbuild_backend"
       backend-path = ["."]
       ```

1. Create the custom build backend `_docbuild_backend.py`:

   ```python
   import shutil
   import re
   from packaging import version
    
   def check_system_requirements():
        config = get_tool_config()
        requirements = config.get("system-requirements", [])
    
        for req in requirements:
            # Regex to split name from optional operator and version
            # e.g., "jing >= 20220510" -> ("jing", ">=", "20220510")
            # e.g., "xmllint" -> ("xmllint", None, None)
            match = re.match(r"^([a-zA-Z0-9_-]+)(?:\s*(>=|==|>)\s*([\d\.]+))?$", req.strip())
            
            if not match:
                print(f"Warning: Could not parse requirement string: {req}")
                continue
    
            name, op, min_v = match.groups()
            
            # 1. Basic check: Does the tool exist?
            if shutil.which(name) is None:
                raise RuntimeError(f"Required system tool '{name}' is not installed or not in PATH.")
    
            # 2. Optional check: Is there a version requirement?
            if op and min_v:
                current_v = get_binary_version(name) # Using our previous helper
                if current_v:
                    if op == ">=" and version.parse(current_v) < version.parse(min_v):
                        raise RuntimeError(f"{name} version {current_v} is too old. Need {req}")
                else:
                    print(f"Warning: Could not determine version for {name}. Proceeding anyway...")
   ```

   Maybe there are more fancy ways to compare versions (perhaps python-semver/python-semver?).

   Additionally, you may need to add additional functions to adhere to the PEP 517 spec.

### Relationships

See also #232.

sushant-suse:
Closes #234

### Changes Made

* **Packaging:** Added `packaging` to `pyproject.toml` (and updated `uv.lock`) for robust semantic version comparison.
* **Validation Engine:** Created `src/docbuild/utils/sysdeps.py` to act as the central registry for first-order system dependencies and provide version-checking logic.
* **New CLI Command:** Added `docbuild doctor` (`src/docbuild/cli/cmd_doctor.py`). It uses `rich.table` to print a beautiful, color-coded matrix showing the installed status and version of all required system tools.
* **Fail-Fast Decorator:** Added a strictly-typed `@requires_system_tools` decorator (using `ParamSpec` and `TypeVar` to satisfy Ruff's `ANN401` rule) to fail fast gracefully. 
* **Command Protection:** Applied the decorator to the `build` command so it halts execution immediately if tools like `daps` or `xmllint` are missing, rather than failing deep in the execution tree with cryptic Python tracebacks.
* **Testing:** Added `tests/cli/test_cmd_doctor.py` and verified that the existing `cmd_build` test suite successfully bypasses the decorator without breaking.
* **Changelog:** Added news fragment `234.feature.rst`.

### Self-Review Checklist

- [x] Dependencies are validated at runtime rather than build-time to support wheel installations.
- [x] Semantic versions are correctly compared using `packaging.version`.
- [x] Decorator halts Click commands cleanly without standard Python tracebacks.
- [x] Ruff strict typing rules (`ANN401`) are satisfied without using `Any`.

**Documentation & CI**:

- [x] Existing tests pass and new tests cover the `doctor` command logic.


src/docbuild/cli/cmd_build/__init__.py
```diff
@@ -50,6 +51,7 @@
     callback=validate_doctypes,
 )
 @click.pass_context
+@requires_system_tools(["daps", "xmllint"])
```
tomschr: I'd suggest to create a tuple in `constants.py`. That will avoid duplicating it in case we want to apply it to different commands.

Apart from that... I'm a bit confused. You have the variable/constant `SYSTEM_DEPENDENCIES`. How does it relate to this list? Why not pass a list derived from `SYSTEM_DEPENDENCIES.keys()`?

src/docbuild/cli/cmd_doctor.py
```diff
@@ -0,0 +1,57 @@
...
+from rich.console import Console
+from rich.table import Table
+
+from ..utils.sysdeps import check_dependencies
+
+
+@click.command(name="doctor", help="Check system dependencies required by docbuild.")
+def doctor() -> None:
+    """Check system dependencies and display their status."""
+    console = Console()
```
tomschr, line 13: This line is not an issue and something for a different PR. But I'm wondering if we should create this console object somewhere else and import it. 🤔

src/docbuild/utils/sysdeps.py
```diff
@@ -0,0 +1,193 @@
...
+T = TypeVar("T")
+
+# First-order dependencies required by docbuild
+SYSTEM_DEPENDENCIES = {
+    "jing": ">=20220510",
+    "trang": None,  # Any version
+    "daps": ">=4",
+    "xmllint": None,
+    "xsltproc": None,
+}
```
tomschr: Maybe we should move that into `constants.py`?

tests/cli/test_cmd_doctor.py
```diff
@@ -0,0 +1,53 @@
...
+
+from docbuild.cli.cmd_doctor import doctor
+
+
+def test_doctor_command_success_no_errors():
+    """Test that doctor exits with 0 when all dependencies are met."""
+    runner = CliRunner()
+
+    # Mock the checker engine so we control the output
+    with patch("docbuild.cli.cmd_doctor.check_dependencies") as mock_check:
```
tomschr: I would recommend to use `patch.object`. For VSCode it's easier when you refactor code.

tests/cli/test_cmd_doctor.py
```diff
@@ -0,0 +1,53 @@
...
+        assert result.exit_code == 0
+        assert "mock-tool" in result.output
+        assert "All system dependencies look good!" in result.output
+
+
+def test_doctor_command_fails_on_missing_tool():
+    """Test that doctor exits with 1 when a dependency is missing."""
+    runner = CliRunner()
+
+    with patch("docbuild.cli.cmd_doctor.check_dependencies") as mock_check:
```
tomschr: Same here, `patch.object`.

src/docbuild/utils/sysdeps.py
```diff
@@ -0,0 +1,193 @@
...
+                # 2. Check version if a requirement exists
+                if req:
+                    found_v = get_binary_version(tool_name)
+                    req_match = re.match(r"^\s*(>=|==|>)\s*([\d\.]+)", req)
+
+                    if found_v and req_match:
+                        op, min_v = req_match.groups()
+                        try:
+                            v_found = parse_version(found_v)
+                            v_min = parse_version(min_v)
```
tomschr: You used the `packaging.version.parse_version` function. That's probably fine. Just to let you know, there is maybe another alternative.

If you really need this module for parsing version strings only, maybe use [`semver`](https://github.com/python-semver/python-semver/)? I know this library and the maintainer quite well... 😆

It may be a bit tricky to parse the version to get a `Version` object. But once you managed it, the `Version` class can compare different versions. This would eliminate some comparison code you have and would delegate all parsing, comparison, and version handling to this lib.

Maybe have a look at their [documentation](https://python-semver.readthedocs.io/en/latest/). :wink:

tomschr (changes requested):
Great PR, thank you very much Sushant! 👍

I have a few minor suggestions and a few questions. Nothing big.

src/docbuild/cli/cmd_build/__init__.py
sushant-suse: Great point Toms. As you suggested in your other comment, I've moved `SYSTEM_DEPENDENCIES` entirely into `constants.py`. I also updated the `@requires_system_tools()` decorator so that if no argument is passed, it automatically defaults to checking `SYSTEM_DEPENDENCIES.keys()`. Now `cmd_build` just uses the bare decorator, which keeps it super clean.

src/docbuild/cli/cmd_doctor.py
sushant-suse: I completely agree. I noticed `cmd_cli.py` instantiates its own Console as well. I kept it localized here for now because importing `CONSOLE from cmd_cli.py` causes a circular import (since `cmd_cli` imports the commands). I think extracting it into a central `ui.py` or `console.py` module in a follow-up PR is a great idea.

WDYT?

src/docbuild/utils/sysdeps.py
sushant-suse: Haha great great. I've replaced packaging with `semver` in `pyproject.toml`. I added a tiny `_coerce_semver` helper to handle system binaries that only report single integers (like 4 instead of 4.0.0), and then let the Version class handle all the comparison magic seamlessly.

src/docbuild/cli/cmd_doctor.py
tomschr: If you ask me, I would probably create a different Python file (for example, `console.py` or `ui.py` as you suggested) and import it from there. I think that would be a great addition. This should eliminate the circular imports, right?

Maybe create an issue for that?

src/docbuild/utils/sysdeps.py
```diff
@@ -0,0 +1,186 @@
...
+    return None
+
+
+def _coerce_semver(version_str: str) -> semver.Version:
+    """Convert loose version strings (like '4' or '2.9') to strict SemVer objects."""
+    parts = version_str.split(".")
+    major = int(parts[0]) if len(parts) > 0 else 0
+    minor = int(parts[1]) if len(parts) > 1 else 0
+    patch = int(parts[2]) if len(parts) > 2 else 0
+    return semver.Version(major, minor, patch)
```
tomschr: That's actually the tricky part. The Python `semver` library is quite strict what it considers a valid semver version. It strictly adheres to major, minor, and patch (with optional prerelease and build parts).

If you only have a major part, you don't have to pass minor and patch. It automatically falls back to zero:

```python
>>> semver.Version("2232")
Version(major=2232, minor=0, patch=0, prerelease=None, build=None)
```

You just pass the string and the library does the rest.

However, if your string contains dots, then it expects the full `major.minor.patch`:

```python
>>> semver.Version.parse("3.4.1")
Version(major=3, minor=4, patch=2, prerelease=None, build=None)
>>> semver.Version.parse("3.4")
Traceback
...
ValueError: 3.4 is not valid SemVer string
```

Long story short: I think you should avoid splitting the version manually. Just delegate that to the library. The easiest way would be this:


```suggestion
    try:
       return semver.Version(version)
    except ValueError:
       return semver.Version.parse(version)
```

However, that may not work for versions like "3.4". Perhaps the library should do this and set patch automatically to zero. I should open a new issue. 😉

In the meantime, maybe you need to apply something similar that is described in the section [Dealing with Invalid Versions](https://python-semver.readthedocs.io/en/latest/advanced/deal-with-invalid-versions.html(

src/docbuild/utils/sysdeps.py
```diff
@@ -0,0 +1,186 @@
...
+            v_found = _coerce_semver(found_version)
+            v_min = _coerce_semver(min_v)
+            is_valid = False
+
+            if op == ">=":
+                is_valid = v_found >= v_min
+            elif op == "==":
+                is_valid = v_found == v_min
+            elif op == ">":
+                is_valid = v_found > v_min
```
tomschr: This manual comparison can be avoided. Look into `semver.Version.match` and the section [Comparing Versions through an Expression](https://python-semver.readthedocs.io/en/latest/usage/compare-versions-through-expression.html).

tomschr (commented):
I've added some comments, hope they help. It's about how to convert a version string into a `Version` object and to compare it.

sushant-suse:
> I've added some comments, hope they help. It's about how to convert a version string into a `Version` object and to compare it.

You're totally right, I was completely reinventing the wheel there. It makes perfect sense to let the library do the heavy lifting. I've rewritten the coercion function to use the native constructor for single-digit versions and `.parse()` with zero-padding for the incomplete ones. I also completely ripped out the manual `if/elif` operator comparisons and replaced them with `v_found.match(expr)` it is so much cleaner! Thanks for pointing out those advanced docs, the `.match` method is fantastic.

src/docbuild/utils/sysdeps.py
```diff
@@ -0,0 +1,189 @@
...
+        pass
+
+    return None
+
+
+def _coerce_semver(version_str: str) -> semver.Version:
+    """Convert loose version strings to strict SemVer objects."""
+    # Handle pure integers (e.g., "4") via the native class constructor
+    if version_str.isdigit():
+        return semver.Version(int(version_str))
```
tomschr: This is not really needed, the class does the conversion automatically:


```suggestion
        return semver.Version(version_str)
```

src/docbuild/utils/sysdeps.py
```diff
@@ -0,0 +1,189 @@
...
+                "found": found_version or "Unknown",
+                "is_installed": True,
+                "is_valid": True,
+                "message": "OK",
+            })
+            continue
+
+        # Parse requirement (e.g. ">=20220510")
+        req_match = re.match(r"^\s*(>=|==|>)\s*([\d\.]+)", requirement)
+        if not req_match:
```
tomschr: This is technically correct. However, I see two problems here:

1. The regex is potentially buggy and restrictive.

   Your regex fails when the version is something like `v4`. On the other side, it matches something like `2..3`. Sooo... maybe we shouldn't try to parse a version. Perhaps it's enough to just extract everything after the operator:

   ```python
   req = re.compile(r"^\s*(>=|<=|==|>|<)\s*(.*)$")
   ```

1. Compile the regex before using it.

   As we are inside a for loop the `re` module has to parse the expression all the time. As we have very few entries that may not a problem. But as soon as we add more, it's a slight speed penalty. Maybe compile the expression _before_ the for loop:

   ```python
   req = re.compile(r"^\s*(>=|<=|==|>|<)\s*(.*)$")
   ```

Then you can change this line to this:



This would use the pre-compiled regex. It's a small optimization step.

```suggestion
        req_match = req.match(requirement)
        if not req_match:
```

tomschr (changes requested):
Thanks Sushant! Sorry, for the additional comments. After I looked more closely for the regex, I think this is an implementation issue.

src/docbuild/utils/sysdeps.py
sushant-suse: I was getting Pylance error, added `#type:ignore`

tomschr (approved):
Thanks Sushant! Awesome work. Just squash-merge it as usual. 👍

Issue #307, opened 2026-05-22
Refactor functionality in src/docbuild/cli/


tomschr:
### Situation

Some CLI commands in `src/docbuild/cli/` mixes CLI functionalities with implementation. IMHO, I think this is wrong and we should provide a cleaner distinction.

Additionally, many files in this directory are named `process.py` which doesn't really reflect the intention.

### Use Case

* Better separate concerns
* Clearer naming

### Possible Implementation

The idea is to separate CLI from the implementation:

* [x] Keep the CLI parts, but import the implementation from a different package.
* [x] Move `process.py` or any other files that are related to implementation into a new directory. Maybe `src/docbuild/tasks`?
* [x] Perhaps split the file(s), create additional subdirectories to make the intention obvious.
* [x] Adjust the test suite

sushant-suse:
I completely agree with this separation of concerns. Keeping the CLI layer strictly for argument parsing/terminal UI and moving the actual logic into pure Python functions in a `tasks/` module will make the core logic much easier to unit test.

Since this is a heavy structural change that will require updating the test suite and moving a lot of files, I am holding off on doing a massive refactor until we discuss this personally.

**Proposal**:

Instead of a giant PR that moves everything at once, we can pick just one simple command, extract its `process.py` logic into `src/docbuild/tasks/`, and use that as a Proof of Concept. Once we are both happy with where the boundary between CLI and implementation sits for that one command, we can roll out the pattern to the rest of the CLI.

tomschr:
Thank you for your comment, Sushant! 👍

Yes, we don't need to refactor that in one big issue. It's not a high priority, but we should tackle that soon when we reach the building part. There it would be necessary to use all the different pieces.

I like your step-by-step approach. 👍

PR #398, opened 2026-07-26
refactor: separate CLI and task logic in cmd_check (PoC for #307)

sushant-suse:
**What was done:**
* **Created `src/docbuild/tasks/` module:** Introduced a dedicated directory for pure Python implementation logic.
* **Refactored `cmd_check`:** Moved the implementation out of `cli/cmd_check/process.py` and into `tasks/check.py`.
* **Decoupled from Click:** The core function `check_repository_files` no longer requires `DocBuildContext`. It now strictly accepts standard Python types (`Path`, `Sequence[Doctype]`), making it completely agnostic to the terminal UI.
* **Thinned the CLI wrapper:** `cli/cmd_check/__init__.py` is now purely responsible for argument parsing, extracting config paths, calling the task, and handling `ClickExceptions`.
* **Updated Tests:** Migrated tests to `tests/tasks/test_check.py`. Testing is now much simpler and cleaner since we no longer need to mock the Click context.

**What is left (Not in this PR):**
* Migrating the remaining commands (`cmd_repo`, `cmd_metadata`, `cmd_build`, `cmd_portal`, etc.).
* Removing the other `process.py` files scattered throughout the CLI module.

If this exact boundary between the CLI and the `tasks/` module looks good, I will use this as the blueprint for subsequent PRs to migrate the rest of the commands!

### Related Issue

#307 

### Checklist
- [x] Tests have been updated and are passing
- [x] A changelog entry (`changelog.d/307.misc.rst`) has been added


tomschr (approved):
Thanks for the PR. Great work, just fix the Ruff error and merge it. 👍

Issue #386, opened 2026-07-23
Command `docbuild portal list` shows unnamed-deliverable for translations


tomschr:
### Project Version

0.21.0

### Python Version

3.12, 3.13, 3.14

### Operating System

Linux

### Steps to Reproduce

We assume, under `~/.config/docbuild` lies all the config files. If they are not, copy it from the repo:

```shell
mkdir -p ~/.config/docbuild/config.d && cp -a src/docbuild/config/xml/data/config.d/* ~/.config/docbuild/config.d/
```

1. Show SLES 16.0 English content:

   ```shell
   $ docbuild portal list 'sles/16.0/en-us'
   en-us/
   └── sles
       └── 16.0
           ├── rn.sles.16.0.16.0-index.release-notes (Prebuilt)
           │   └── URL: /releasenotes/sles/16.0/index.html
           ├── sles.16.0.dc-sap-saptune (DC-SAP-saptune)
           ├── [...]
           └── sles.16.0.dc-sles-x86-64-agama-based-installation (DC-SLES-x86-64-agama-based-installation)
   ```

   => You get a long list with names, IDs etc.

1. Show the same for German.

   => You will get "unnamed-deliverable"

### Actual Behavior

```shell
$ docbuild portal list 'sles/16.0/de-de'
de-de/
└── sles
    └── 16.0
        ├── unnamed-deliverable
        ├── unnamed-deliverable
        ├── unnamed-deliverable
        ├── unnamed-deliverable
        ├── unnamed-deliverable
        ├── unnamed-deliverable
        ├── unnamed-deliverable
        ├── unnamed-deliverable
        ├── unnamed-deliverable
        ├── unnamed-deliverable
        ├── unnamed-deliverable
        ├── unnamed-deliverable
        ├── unnamed-deliverable
        ├── unnamed-deliverable
        ├── unnamed-deliverable
        ├── unnamed-deliverable
        ├── unnamed-deliverable
        ├── unnamed-deliverable
        ├── unnamed-deliverable
        └── unnamed-deliverable
```

=> It shows the correct amount of lines, but not the names.

### Expected Behavior

I would have expect that the command shows DC names and IDs like in the English case. After all it's a deliverable, but a reference like this:

```xml
<locale lang="de-de" sitemap="false">
   <branch>maintenance/SLE16.0</branch>
   <subdir>l10n/de-de</subdir>
   <!--************************ about ************************-->
   <deliverable type="ref">
[...]

PR #400, opened 2026-07-27
Fix #386 `docbuild portal list` to fix "unnamed-deliverable"

tomschr:
The command `docbuild portal list` shows `unnamed-deliverable` for translations.

This fix makes the following changes:

* Introduces the `locale_en` cached property to return the English locale. This is used in `dcfile` and `deliverableid`.
* The property `dcfile` considers translations (a deliverable with `<ref/>`) now.
* Introduces `deliverableid` to return the correct ID for a translation.

### DIFF
--- a/changelog.d/400.bugfix.rst
+++ b/changelog.d/400.bugfix.rst
@@ -1,0 +1,1 @@
+Correct :command:`docbuild portal list` to fix "unnamed-deliverable" output for translated deliverables. The deliverable ID is now correctly retrieved from the English locale node when the current node is a reference deliverable.
--- a/src/docbuild/cli/cmd_portal/cmd_list.py
+++ b/src/docbuild/cli/cmd_portal/cmd_list.py
@@ -115,13 +115,13 @@ def build_deliverable_branch(
     repo_format: str | None,
 ) -> None:
     """Format and append a single deliverable node to the Rich Tree."""
-    d_id = deliv.xml.node.get("id", "unnamed-deliverable")
+    d_id = deliv.xml.deliverableid or "unnamed-deliverable"
 
     # 1. Format the main display name
     display_name = get_display_name(deliv, d_id)
     deliv_branch = docset_branch.add(display_name)
 
-    # 2. Automatically show URLs for prebuilt deliverables
+    # 2a. Automatically show URLs for prebuilt deliverables
     if deliv.xml.is_prebuilt:
         for url_node in deliv.xml.node.xpath("prebuilt/url"):
             if href := url_node.get("href"):
--- a/src/docbuild/models/deliverable/view.py
+++ b/src/docbuild/models/deliverable/view.py
@@ -64,6 +64,11 @@ class DeliverableXMLView:
         return self.docset_node.attrib.get("path", None)
 
     @cached_property
+    def locale_en(self) -> etree._Element | None:
+        """Return the English locale node (``<locale lang="en-us">``) or None if absent."""
+        return self.docset_node.find("resources/locale[@lang='en-us']")
+
+    @cached_property
     def lang(self) -> LanguageCode:
         """Return the language and country code (e.g. ``'en-us'``)."""
         return LanguageCode(language=self.node.getparent().attrib.get("lang").strip())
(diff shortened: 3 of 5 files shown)


sushant-suse:
Hi Toms. Thanks for the PR. Awesome work as always. I pulled down the branch and ran the test suite, everything is passing perfectly on my end, and the tests you added cover the exact scenarios flawlessly. Great catch on the `<ref>` tag missing the local `@id` and `dcfile`!
I just have one minor suggestion for the implementation in `view.py`:
In `dcfile` and `deliverableid`, we are currently doing this:
```python
dcnode = cast(etree._Element, self.locale_en).find(...)
```


While `cast()` keeps Pylance happy, it might be slightly safer to use a strict type guard in case the `en-us` locale is somehow missing from a malformed XML file (which would cause `.find()` to throw an `AttributeError` on `None`).
Something like this:
```python
if self.locale_en is not None:
    dcnode = self.locale_en.find(f"deliverable[@id={refid!r}]/dc")
    if dcnode is not None:
        return dcnode.attrib.get("file")

```

Other than that tiny guard suggestion, this looks completely solid and is good to go!

sushant-suse (approved):

tomschr:
Great observation! :+1:

> [...]
> ```python
> dcnode = cast(etree._Element, self.locale_en).find(...)
> ```
> 
> While `cast()` keeps Pylance happy, it might be slightly safer to use a strict type guard in case the `en-us` locale is somehow missing from a malformed XML file (which would cause `.find()` to throw an `AttributeError` on `None`). Something like this:
> 
> ```python
> if self.locale_en is not None:
>     dcnode = self.locale_en.find(f"deliverable[@id={refid!r}]/dc")
>     if dcnode is not None:
>         return dcnode.attrib.get("file")
> ```
> 
> Other than that tiny guard suggestion, this looks completely solid and is good to go!

That is technically absolutely correct. However, I did that on purpose. :wink:

Remember, we rely on validation. Before we process the portal config, the XML structure is validated against the RELAX NG schema with `jing`.

That means, when the portal config is invalid, the script will not continue. But if it does, the structure is guaranteed.

As a consequence, because of the validation we can avoid some checks. For example, because a `<locale>` element is required we don't need to explicitly check if it exists. This keeps the code clean, readable, and we don't have to test all the if-clauses to reach good coverage. We can focus on the necessary checks. Therefor I think, in this case it's not needed as we rely on validation.

Or do I miss something? Does my logic contain a flaw? Let me know!

sushant-suse:
> Great observation! 👍
> 
> > [...]
> > ```python
> > dcnode = cast(etree._Element, self.locale_en).find(...)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > While `cast()` keeps Pylance happy, it might be slightly safer to use a strict type guard in case the `en-us` locale is somehow missing from a malformed XML file (which would cause `.find()` to throw an `AttributeError` on `None`). Something like this:
> > ```python
> > if self.locale_en is not None:
> >     dcnode = self.locale_en.find(f"deliverable[@id={refid!r}]/dc")
> >     if dcnode is not None:
> >         return dcnode.attrib.get("file")
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Other than that tiny guard suggestion, this looks completely solid and is good to go!
> 
> That is technically absolutely correct. However, I did that on purpose. 😉
> 
> Remember, we rely on validation. Before we process the portal config, the XML structure is validated against the RELAX NG schema with `jing`.
> 
> That means, when the portal config is invalid, the script will not continue. But if it does, the structure is guaranteed.
> 
> As a consequence, because of the validation we can avoid some checks. For example, because a `<locale>` element is required we don't need to explicitly check if it exists. This keeps the code clean, readable, and we don't have to test all the if-clauses to reach good coverage. I think, it's unnecessary if we have already validation.
> 
> Or do I miss something? Does my logic contain a flaw? Let me know!

You are absolutely right, I had temporarily forgotten about the strict jing RELAX NG validation.
There is no flaw in your logic at all. I've approved the PR. As always, great work, please go ahead and merge it. Thanks!

PR #401, opened 2026-07-28
refactor: separate CLI and task logic in cmd_repo (Relates #307)

sushant-suse:
Following up on the architectural pattern established for `cmd_check`, this PR applies the same decoupling to the `cmd_repo` commands. 

**What was done:**
* **Extracted logic:** Moved the implementation out of `cli/cmd_repo/process.py` and into `tasks/repo.py` (`clone_repositories`).
* **Decoupled from Click:** The core task no longer depends on `DocBuildContext`. It now strictly takes standard Python types (`Path`, `tuple`).
* **Cleaned the CLI wrappers:** Updated `cmd_clone.py` to purely handle argument extraction and delegate to the new task. (`cmd_list` and `cmd_dir` required no changes as they just perform simple path lookups).
* **Modernized Tests:** Replaced the bloated, CLI-coupled tests in `tests/cli/cmd_repo/test_process.py` with clean unit tests in `tests/tasks/test_repo.py`. Added a lightweight delegation test to `test_cmd_clone.py`.

### Related Issue

Relates to #307 

### Checklist

- [x] A changelog entry was already added in the previous PoC PR


src/docbuild/tasks/repo.py
```diff
@@ -1,34 +1,30 @@
...
+    repos: tuple[str, ...]
+) -> int:
     """Process the cloning of repositories.
 
-    :param context: The DocBuildContext object containing configuration.
+    :param main_portal_config: Path to the main portal config XML.
+    :param repo_dir: Path to the directory where bare repos should be stored.
     :param repos: A tuple of repository selectors. If empty, all repos are used.
-    :return: An integer exit code.
-    :raises ValueError: If configuration paths are missing.
```
tomschr, line 23: Should we keep this line?

tomschr (commented):
Great work! 👍

I just have one minor question (see below). Otherwise merge it. 👍

src/docbuild/tasks/repo.py
sushant-suse: Great catch Toms. I actually removed the `:raises ValueError:` line intentionally because of how the new architecture works. Since `clone_repositories` now strictly accepts `Path` objects as arguments, it no longer digs into the context or checks for missing configuration itself. We moved that validation up to the CLI wrapper (`cmd_clone.py`), which now explicitly raises a `click.ClickException` if `envconfig` is missing *before* the task is even called. Since the task itself no longer raises that `ValueError`, I dropped it from the docstring. I will go ahead and merge this now based on your approval, but let me know if you think I should document it differently in the future.

tomschr: Ahh right. When you look at a PR, the context is sometimes a bit limited.

I'm fine of what you did. Merge it. :+1:

tomschr (approved):

PR #402, opened 2026-07-29
refactor: separate CLI and task logic in cmd_portal (Relates #307)

sushant-suse:
**What was done:**

* **Extracted logic:** Moved the XML parsing and Jing validation logic out of `cli/cmd_portal/process.py` and into `tasks/portal.py`.
* **Decoupled from Click:** The core validation tasks no longer accept the bloated `DocBuildContext` object. They now take standard Python types (`Path`, `verbose: int`, etc.), making them highly reusable.
* **Cleaned CLI wrapper:** `cmd_validate.py` now purely handles extracting paths from the Click context/flags and delegates everything else to `validate_portal_config`.
* **Cross-module fixes:** Updated imports in `cmd_metadata` and `cmd_check` that were relying on the old `process.py` file.
* **Modernized Tests:** Safely migrated all the heavy business-logic and validation tests to `tests/tasks/test_portal.py`. Rewrote `tests/cli/cmd_portal/validate/test_cmd_portal.py` to exclusively test the Click CLI wrapper behavior. 

### Related Issue

Relates to #307 

### Checklist

- [x] A changelog entry was already added in the previous PoC PR


src/docbuild/cli/cmd_portal/cmd_validate.py
```diff
@@ -44,26 +44,26 @@ def validate(
...
-    configdir_path = Path(configdir).expanduser()
+    # Type guard: Ensure envconfig is loaded before accessing its attributes
+    if context.envconfig is None:
+        raise click.ClickException("Environment configuration is missing.")
+
     if not main_portal_config:
-        main_portal_config = env.paths.main_portal_config.expanduser()
+        main_portal_config = Path(context.envconfig.paths.main_portal_config).expanduser()
     else:
         main_portal_config = Path(main_portal_config).expanduser()
```
tomschr: Is this necessary? I think Pydantic is taking care of expanding the user and having a `Path` object. This would simplify it like this:


```suggestion
        main_portal_config = context.envconfig.paths.main_portal_config
```

I've tested it and it works without the `else` clause.

src/docbuild/cli/cmd_portal/cmd_validate.py
```diff
@@ -44,26 +44,26 @@ def validate(
...
 
     if not portal_schema:
-        portal_schema = env.paths.portal_rncschema.expanduser()
+        portal_schema = Path(context.envconfig.paths.portal_rncschema).expanduser()
     else:
         portal_schema = Path(portal_schema).expanduser()
```
tomschr: Same here:


```suggestion
        portal_schema = context.envconfig.paths.portal_rncschema
```

src/docbuild/tasks/portal.py
```diff
@@ -219,12 +217,12 @@ async def parse_portal_config(filepath: Path | str) -> etree._ElementTree:
...
 async def run_checks_and_display(
     tree: etree._ElementTree,
-    context: DocBuildContext,
+    verbose: int = 0,
 ) -> bool:
     """Execute registered Python checks and print formatted results.
 
     :param tree: Parsed XML tree to check.
-    :param context: :class:`DocBuildContext` used to read verbosity.
+    :param verbose: Verbosity level (0=normal, 1=verbose, 2=debug).
```
tomschr: I think it's misleading to say 0=normal. In most cases, having zero verbosity at the CLI level is having disabled. Or am I mislead?

tests/tasks/test_portal.py
```diff
@@ -0,0 +1,403 @@
...
+@patch.object(portal_task.etree, "parse", side_effect=ValueError("Generic test error"))
+async def test_validate_portal_config_generic_parsing_error(
+    mock_etree_parse, mock_run_validation, tmp_path
+):
+    mock_run_validation.return_value = ValidationResult(True, 0, "")
+    xml_file = tmp_path / "file.xml"
+    xml_file.touch()
+    schema = tmp_path / "schema.rnc"
+
+    with pytest.raises(ValueError, match="Generic test error"):
```
tomschr: The "Generic test error" is here at two locations. To not repeat yourself, use the one that you passed in `patch.object`:

```suggestion
    with pytest.raises(ValueError, match=str(mock_etree_parse.side_effect)):
```

Tested it and it works.

src/docbuild/tasks/check.py
tomschr, line 1: According to the coverage report, these are not covered by tests:

```
src/docbuild/tasks/check.py                       53      0     22      1  98.7%   65->69
```

tomschr (changes requested):
Great work! 👍 Just a few comments.

src/docbuild/cli/cmd_portal/cmd_validate.py
sushant-suse: Great catch Toms. You're completely right. Since Pydantic and Click (via `path_type=Path`) are already handling the path resolution for us, wrapping it again was completely redundant. I have simplified both of those blocks.

src/docbuild/tasks/portal.py
sushant-suse: Yes, 'disabled' is definitely more accurate for a CLI tool context.

src/docbuild/tasks/check.py
sushant-suse: Updated the tests

tomschr (approved):
Looks good! Great work! Let's merge it. 👍

PR #284, opened 2026-05-18
Adjust docbuild metadata

tomschr:
Implement translation-aware metadata pipeline with localized date parsing

Includes:
* Add DAPS execution helper and metadata pipeline refactors
* Support translation handling and json cache paths
* Add localized date parsing heuristics
* Apply code quality suggestions, refactors and code quality suggestions
* Add DAPS metadata runner and translation-aware processing
* Merge static + dynamic metadata into compiled manifests
* Add json cache paths and localized date parsing heuristics

PR #409, opened 2026-07-30
Fix 268: Implement dobuild metadata

tomschr:
Replaces #284 

This PR contains:

* `src/docbuild/cli/defaults.py`:
  Add missing `builddir`, `dcfile`, and `output` in key `build.daps.meta` of `DEFAULT_ENV_CONFIG`.

* `src/docbuild/logging.py`: 
  Add task name as logger output

* Implement the metadata extraction in `src/docbuild/tasks/metadata`

* Use semaphore in runner to speed up things. Without it, sles/16.0 needed 766s to process metadata. With the semaphore it's only 140s.

* `src/docbuild/models/manifest.py`:
  * Fix the Description model; it wrongly assumed `<desc>` elements but we need now `descriptions/desc`.
  * The `Category.from_xml_node` uses `categories/category` and avoids `category`. Additionally, yields the correct instances for each unique category ID.
  * The `SingleDocument` received `default`, and `model_config` attributes. The argumetn for `datemodified` field were changed from `serialization_alias` to `alias` as its value didn't appear in the JSON output.
  * The `Document` received `category`. It's excluded in the JSON output if it's `None` (=not needed).

* `src/docbuild/models/doctype.py`:
  Had to make Doctype a bit more flexible:
  * Make `language` optional, and set default to `en-us`. This helps to reduce typing and use an abbreviated form of `sles/16.0` to mean the English deliverables in that set. 
  * In `Doctype.from_str`, it contains a new argument `default_lang` (set to "en-us" by default). This helps to select `*` for all languages.

* Updated the autodoc API files and created a new "Creating Metadata" topic.
  Ignore the files in `docs/source/reference/_autoapi/`

----------

```shell
$ docbuild -vvv metadata 'sles/16.0'
[...]
  | DC-SLES-networkmanager-nmcli [en-us]
  | DC-SLES-ulp [en-us]
  | DC-SLES-ppc-agama-based-installation [ja-jp]
  | DC-SLES-ntp-time-synchronization [ja-jp]
  | DC-SLES-systemd-basics [en-us]
  | DC-SLES-ansible-roles [en-us]
 > Result: /home/toms/.cache/docbuild/default-env/meta/sles/16.0.json
Elapsed time xxx
```


sushant-suse:
Hi @tomschr, this is a fantastic update. The speedup from 766s to 140s is incredible, and the `asyncio.Semaphore` is the perfect tool for the job. The Pydantic model fixes for `Category` and `Document` (schema v7 alignment) look spot on, and making the language optional in `Doctype` is a huge DX improvement.

**The catch:**
I created a PR for the issue #307 into `main`, which completely decoupled the metadata logic from the CLI. I deleted `metaprocess.py` and extracted the purely functional logic into `src/docbuild/tasks/metadata.py`. Because it removed `DocBuildContext` from the task signatures (passing explicit `Path` objects instead), you are going to hit some hefty merge conflicts when you rebase against `main`.

The good news is that your new `Semaphore` logic slides perfectly into the decoupled architecture. Once you pull `main`, you can adapt your concurrency logic into the new `tasks/metadata.py` file like this:

```python
    # Inside tasks/metadata.py -> process_doctype()
    worker_limit = max(1, min(max_workers, len(deliverables)))
    semaphore = asyncio.Semaphore(worker_limit)

    async def process_deliverable_limited(d: Deliverable) -> tuple[bool, Deliverable]:
        async with semaphore:
            # Using the decoupled signature!
            return await process_deliverable(
                d, repo_dir, tmp_repo_dir, meta_cache_dir, dapsmetatmpl
            )

    tasks = [
        asyncio.create_task(
            process_deliverable_limited(d),
            name=f"process_deliverable_{d.full_id}",
        )
        for d in deliverables
    ]
```
 
 
Your Pydantic model updates (`manifest.py` and `doctype.py`), logging updates, and defaults should merge cleanly. For the task logic, we just need to weave your `Semaphore` into the new decoupled `metadata.py` file. Would you like to handle the rebase yourself, or would you like me to push a commit to this branch that resolves the architecture conflicts for you? Awesome work on this 🚀

sushant-suse (approved):

PR #411, opened 2026-07-31
refactor: separate CLI and task logic in cmd_metadata (Resolves #307)

### LINKED ISSUE #307: Refactor functionality in src/docbuild/cli/

sushant-suse:
This is the final PR for Issue #307

**What was done:**

* **Extracted logic:** Moved all the XML parsing, `daps` subprocess execution, repository handling, and JSON merging from `cli/cmd_metadata/metaprocess.py` to a new `tasks/metadata.py` module.
* **Decoupled from Click:** The tasks no longer accept the bloated `DocBuildContext` object. They now take standard `Path` objects, making them purely functional and highly reusable.
* **Cleaned CLI wrapper:** `cli/cmd_metadata/__init__.py` purely handles unpacking the Pydantic variables into `Path` objects and passing them to `generate_metadata`.
* **Cross-module fixes:** Updated imports in `cmd_check` to reference the new task module.
* **Modernized Tests:** Safely migrated ~30KB of business-logic tests into `tests/tasks/test_metadata.py`. Created a lightweight Click testing file specifically for `cli/cmd_metadata`.
* Fixed a couple of minor Pydantic typing bugs that surfaced during the extraction.

### Related Issue

Closes #307 

### Checklist
- [x] A changelog entry was already added in the first PoC PR


src/docbuild/cli/cmd_metadata/__init__.py
```diff
@@ -51,18 +52,41 @@ def metadata(
...
-                    context,
-                    doctypes,
+                    main_portal_config=main_portal_config,
+                    tmp_metadata_dir=tmp_metadata_dir,
+                    repo_dir=repo_dir,
+                    tmp_repo_dir=tmp_repo_dir,
+                    meta_cache_dir=meta_cache_dir,
+                    dapsmetatmpl=dapsmetatmpl,
+                    max_workers=max_workers,
+                    doctypes=list(doctypes),
```
tomschr, line 91: I'm afraid, but the signature of this function is like this:

```python
def process(
    context: DocBuildContext,
    doctypes: Sequence[Doctype] | None,
    *,
    exitfirst: bool = False,
    skip_repo_update: bool = False
) -> CoroutineType[Any, Any, int]
```

We could change that signature.

tomschr (changes requested):
Thanks for the PR! 👍

When I try to run it, I get this error:

```
$ docbuild -vvv metadata 'trd/wordpress'
Config path: /home/toms/.config/docbuild/config.d
Elapsed time 0.00s
[...]
  File "/home/toms/repos/GH/opensuse/docbuild.git/main/.venv/lib/python3.14/site-packages/click/decorators.py", line 34, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "/home/toms/repos/GH/opensuse/docbuild.git/main/src/docbuild/cli/cmd_metadata/__init__.py", line 81, in metadata
    process(
    ~~~~~~~^
        main_portal_config=main_portal_config,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    ...<8 lines>...
        skip_repo_update=skip_repo_update,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
TypeError: process() got an unexpected keyword argument 'main_portal_config'
```

It's about the signature of the function. We need to adjust `process`.

src/docbuild/cli/cmd_metadata/__init__.py
tomschr: Hmn, when I look into `process`, changing the signature is harder than I thought. :thinking:

It's because the functions `process_doctype` and `store_productdocset_json` are called and require the `context` to be passed. If we change the `process` and get rid of `context` as argument, we need to change the others as well.

Maybe it would be beneficial as we list only those arguments that we really require. And we don't have to rely on the context. Kind of least principle of what we really need.

What do you think?

sushant-suse:
> Thanks for the PR! 👍
> 
> When I try to run it, I get this error:
> 
> ```
> $ docbuild -vvv metadata 'trd/wordpress'
> Config path: /home/toms/.config/docbuild/config.d
> Elapsed time 0.00s
> Traceback (most recent call last):
> [...]
>   File "/home/toms/repos/GH/opensuse/docbuild.git/main/.venv/lib/python3.14/site-packages/click/decorators.py", line 34, in new_func
>     return f(get_current_context(), *args, **kwargs)
>   File "/home/toms/repos/GH/opensuse/docbuild.git/main/src/docbuild/cli/cmd_metadata/__init__.py", line 81, in metadata
>     process(
>     ~~~~~~~^
>         main_portal_config=main_portal_config,
>         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>     ...<8 lines>...
>         skip_repo_update=skip_repo_update,
>         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>     )
>     ^
> TypeError: process() got an unexpected keyword argument 'main_portal_config'
> ```
> 
> It's about the signature of the function. We need to adjust `process`.

Hi Toms, thanks for the review. Since our branches crossed paths, my CLI wrapper here is passing the explicit arguments, but it's calling your new `runner.py` which is still expecting `DocBuildContext`. I completely agree with your thinking. I'm going to update `runner.py`, `daps.py`, and `manifest.py` right now to drop `context` and accept these explicit arguments instead.

src/docbuild/cli/cmd_metadata/__init__.py
sushant-suse: I'll update `process`, `process_doctype`, `store_productdocset_json`, and the underlying daps functions right now to drop `context` and accept these explicit arguments instead. I'll get those updated and push the fix shortly.

tomschr (approved):
Thanks, awesome work! 👍 We can merge it.

PR #413, opened 2026-08-02
Change dependency group "devel" -> "dev"

tomschr:
This PR renames the `devel` group dependency name to `dev`.

### Reason
This makes it a bit easier with `uv` as you don't have to select this group explicitly anymore. The uv tool [syncs the `dev` group by default](https://docs.astral.sh/uv/concepts/projects/dependencies/#default-groups).

### Changes
* Change the name in `pyproject.toml`. Keep the name `devel` as a compatibility layer; it's just an alias to `dev` now.
* Update `uv.lock`
* Update the documentation

### See also

* https://docs.astral.sh/uv/concepts/projects/dependencies/#development-dependencies
* https://docs.astral.sh/uv/reference/cli/#uv-add--dev
* https://docs.astral.sh/uv/reference/environment/#uv_dev

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -20,7 +20,7 @@ Docbuild builds documentation from DocBook 5/ASCIIDoc, manages XML configs, clon
 ## Setup
 
 1. Create a virtual environment with `uv venv`
-2. Install dependencies: `uv sync --frozen --no-sync --group devel`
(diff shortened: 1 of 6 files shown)


sushant-suse (approved):

sushant-suse:
Great work Toms. Please go ahead and merge it.

PR #416, opened 2026-08-05
Fix description logic in output JSON

tomschr:
The "global" and "local" descriptions need to be merged in the correct way.

* `src/docbuild/tasks/metadata/manifest.py`
   Take into account the "treatment" attribute

### DIFF
--- a/changelog.d/416.bugfix.rst
+++ b/changelog.d/416.bugfix.rst
@@ -1,0 +1,1 @@
+Fix description logic in output JSON and how global and local descriptions are merged.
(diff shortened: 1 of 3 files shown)


sushant-suse (approved):
Hi Toms, this is an awesome addition.

The merging logic in `merge_descriptions_with_treatment` is super clean, and using `.model_copy(deep=True)` is the perfect way to ensure we don't accidentally mutate shared description objects across different scopes. I also noticed you already used the new decoupled signature for `store_productdocset_json` in your tests, thanks for syncing up with the recent architecture changes so quickly. This looks good to merge from my end.

PR #419, opened 2026-08-05
Fix #418: Improve test for Portal Config schema

### LINKED ISSUE #418: Improve test cases for Portal Config schema
### Situation

Inside `tests/config/xml/schema_cases/`, we have only a very rudimentary XML file. As the Portal Config schema is complex, we need to test more structures.

### Use Case / User Story

* Add invalid and valid XML files
* Add small files that test a specific part of the config (for example, languages, categories, references etc.)

### Possible Implementation

Add the necessary test XML file, for example:

* Categories and their translations
* Product families and series
* Locale entries
* References from a translation to the English original text

Test it with:

```
$ source devel/activate-aliases.sh
$ upytest -v tests/config/xml/test_schema.py
```

tomschr:
This PR contains invalid and valid test cases of Portal Config XML files.

### DIFF
--- a/changelog.d/419.feature.rst
+++ b/changelog.d/419.feature.rst
@@ -1,0 +1,1 @@
+Added more test cases for the Portal Config schema.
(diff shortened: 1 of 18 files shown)


tomschr:
I think, it's time to merge it even if it's incomplete.

PR #426, opened 2026-08-06
Optimise metadata by grouping deliverables

tomschr:
Optimise `docbuild metadata` by grouping deliverables that share the same repository and branch into a single shared `git worktree`. This reduces the number of worktree checkout operations from one per deliverable to one per unique `(repository, branch)` pair, significantly reducing overhead when many DC files live on the same branch.

This PR may lead to nowhere, depending on the outcome.


tomschr:
I think, this contains too many conflicts as there were many changes recently. So the best would be probably to close this and restart again, if really needed.

PR #427, opened 2026-08-07
Refactor Product enum

tomschr:
The `Product` class needed a refactoring attempt. I had always some type problems in my VS Code editor which was annoying.

I think the refactoring created a more static class which is easier to handle for editors (and probably type checkers) than the dynamic one.

Changes in this PR:
* Remove obsolete `VALID_PRODUCTS` and `ALLOWED_PRODUCTS`, not really needed in the code anywhere.
* Replace dynamic Product enum construction with a direct static `Product` class.
* Store full product names as Product values and add `Product.acronym` for canonical short IDs.
* Extend `Product._missing_` to accept acronym input and wildcard (*) before falling back to base error handling.
* Make `_missing_` typing LSP-compatible by using `Self | None` in base and subclass definitions.
* Fix `StrEnumMeta.__getitem__` typing/signature issues (parameter naming and safe typing casts).
* Update `Product` docstrings to clarify full-name values, acronym semantics, and why `_missing_` overrides base behavior.

### DIFF
--- a/src/docbuild/cli/cmd_portal/cmd_list.py
+++ b/src/docbuild/cli/cmd_portal/cmd_list.py
@@ -8,11 +8,12 @@ from lxml import etree  # type: ignore
 from rich.console import Console
 from rich.tree import Tree
 
-from docbuild.cli.context import DocBuildContext
-from docbuild.config.xml.list import list_all_deliverables
-from docbuild.models.deliverable import Deliverable
-from docbuild.models.doctype import Doctype
-from docbuild.tasks.portal import parse_portal_config
+from ...cli.context import DocBuildContext
+from ...config.xml.list import list_all_deliverables
+from ...models.deliverable import Deliverable
+from ...models.doctype import Doctype
+from ...models.product import Product
+from ...tasks.portal import parse_portal_config
 
 
 def build_hierarchy(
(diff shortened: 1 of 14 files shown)


src/docbuild/cli/cmd_portal/cmd_list.py
```diff
@@ -233,8 +233,8 @@ def validate_docsets_against_xml(
     errors = []
 
     for dt in doctypes:
-        if dt.product and "*" not in dt.product and dt.docset and "*" not in dt.docset:
-            prod_val = dt.product.value
+        if dt.product and dt.product != dt.product.ALL and dt.docset and "*" not in dt.docset:
```
sushant-suse: Hi Toms, a nitpick: Using `dt.product.ALL` works because Enums allow member access from instances, but it can read a bit weirdly (like it's an instance property). It might be cleaner and more standard to import `Product` and use `Product.ALL` here instead:
`if dt.product and dt.product != Product.ALL and ...`

It also applied to `src/docbuild/config/xml/list.py` file.

src/docbuild/models/doctype.py
```diff
@@ -278,7 +278,7 @@ def product_xpath_segment(self: Self) -> str:
         Example: "product[@id='sles']" or "product"
         """
         if self.product != Product["ALL"]:
```
sushant-suse: Again a nitpick: Just a few lines above in `docset_xml()`, we check this using `if self.product != Product.ALL:`. We should probably stick to `Product.ALL` here as well just to keep the syntax consistent across the file.

sushant-suse (commented):
Hi Toms, thanks for this great refactor. I just left two minor nitpicks about how we reference the `ALL` enum member for consistency, but otherwise this looks solid.

src/docbuild/cli/cmd_portal/cmd_list.py
tomschr: Ahh, great catch! I'll adjust it accordingly. Thanks! :+1:

src/docbuild/models/doctype.py
tomschr: Good catch! :+1: I also fixed it in commit e7d87838.

PR #428, opened 2026-08-07
Iterate over all possible doctypes

tomschr:
This can be helpful if you need to "dissect" all possible values into distinct Doctypes.

Changes:
* Add `Doctype.iter_doctypes` to iterate over all docset and language combinations.
* Take into account wildcard values (`*`). They can only be expanded, if we pass the portal root tree.
* Amend test case
* Update API documentation

### DIFF
--- a/src/docbuild/models/doctype.py
+++ b/src/docbuild/models/doctype.py
@@ -1,9 +1,12 @@
 """Module for defining the Doctype model."""
 
+from collections.abc import Iterator
+from itertools import product
 import re
 from re import Pattern
-from typing import ClassVar, Self
+from typing import ClassVar, Self, cast
 
+from lxml import etree  # type: ignore
 from pydantic import BaseModel, Field, field_validator
 
 from .language import LanguageCode
@@ -151,7 +154,7 @@ class Doctype(BaseModel):
 
         return all(
             [
-                self.product == other.product or self.product == Product.ALL,
+                self.product == other.product or self.product is Product.ALL,
                 set(other.docset).issubset(self.docset) or "*" in self.docset,
                 other.lifecycle in self.lifecycle,
                 set(other.langs).issubset(self.langs) or "*" in self.langs,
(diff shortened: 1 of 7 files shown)


src/docbuild/models/doctype.py
```diff
@@ -149,7 +152,7 @@ def __contains__(self: Self, other: "Doctype") -> bool:
 
         return all(
             [
-                self.product == other.product or self.product == Product.ALL,
+                self.product == other.product or self.product.value == "*",
```
sushant-suse: Hi Toms, Since you did a refactor of the Product enum in the PR #429, it's safer and more Pythonic to compare against the enum member directly using is.

```suggestion
                self.product == other.product or self.product is Product.ALL,
```

src/docbuild/models/doctype.py
```diff
@@ -166,6 +169,95 @@ def __hash__(self: Self) -> int:
...
+        symbolic.
+
+        >>> doctype = Doctype.from_str("sles/15-SP6,15-SP7/en-us,de-de")
+        >>> [
+        ...     f"{item.product.value}/{item.docset[0]}/{item.langs[0].language}"
+        ...     for item in doctype.iter_doctypes()
+        ... ]
+        ['sles/15-SP6/de-de', 'sles/15-SP6/en-us', 'sles/15-SP7/de-de', 'sles/15-SP7/en-us']
+        """
+        has_product_wildcard = self.product.value == "*"
```
sushant-suse: Same enum consistency here:

```suggestion
        has_product_wildcard = self.product is Product.ALL
```

src/docbuild/models/doctype.py
```diff
@@ -243,7 +335,7 @@ def xpath(self: Self) -> str:
         """
         # Example: /sles/15-SP6@supported/en-us,de-de
         product = "product"
-        if self.product != Product.ALL:
+        if self.product.value != "*":
```
sushant-suse: I would suggest the same above here as well.

src/docbuild/models/doctype.py
```diff
@@ -277,7 +369,7 @@ def product_xpath_segment(self: Self) -> str:
 
         Example: "product[@id='sles']" or "product"
         """
-        if self.product != Product["ALL"]:
+        if self.product.value != "*":
```
sushant-suse: I would suggest the same above here as well.

tests/models/test_doctype.py
```diff
@@ -239,6 +240,123 @@ def test_hash_with_doctype():
...
+
+    assert result == [
+        "sles/15-SP6/de-de",
+        "sles/15-SP6/en-us",
+        "sles/15-SP7/de-de",
+        "sles/15-SP7/en-us",
+    ]
+
+
+def test_iter_wildcard_docset_expands_with_portal_root():
```
sushant-suse, line 272: Hi, can you check the indentation here. It looks like `test_iter_wildcard_docset_expands_with_portal_root`, `test_iter_wildcard_lang_expands_with_portal_root`, and `test_iter_wildcard_remains_symbolic_without_portal_root` all have their bodies indented with 8 spaces instead of the standard 4.

sushant-suse (approved):
Hi Toms, thanks for the great feature. I left a few minor comments about indentation in the tests and keeping our new Product enum usage consistent, but otherwise, this is good to go.

src/docbuild/models/doctype.py
tomschr: Great catch! :+1: I've fixed it in commit 56704060

src/docbuild/models/doctype.py
tomschr: Great catch! :+1: I've fixed it in commit 56704060

tests/models/test_doctype.py
tomschr: Good point! Not sure why I added 8 instead of 4 spaces. :thinking:  I'll make sure to adjust it accordingly. Added in commit 484f2819

src/docbuild/models/doctype.py
tomschr: Thanks, good point! :+1: Committed in 1b3e9852.

src/docbuild/models/doctype.py
tomschr: Thanks, committed in 5670406.

PR #429, opened 2026-08-07
feat: bypass validation for config list command by default

### LINKED ISSUE #384: Change validation behaviour of `config list`
### Situation

When executing `docbuild config list` currently the configuration gets validated before being listed.
This way there is no way to list the current configuration if it does not pass validation.

### Use Case

Especially when debugging an erroneous partial config (#343) it would be useful if `config list` would just list the config without validation.

### Possible Implementation

As there already is the command `config validate` the current behaviour of `config list` can arguably  be dropped entirely or it could be moved to an option `--validate`.

sushant-suse:
Closes #384

**What was done:**
* **Bypassed Validation:** Updated the root `cli()` command to intercept the `config list` invocation. If detected (and `--validate` is omitted), it passes `skip_validation=True` to the config loaders.
* **Stored Raw Configs:** Added `raw_appconfig` and `raw_envconfig` to `DocBuildContext` to safely hold the unvalidated TOML dictionaries.
* **Added `--validate` flag:** Added a new boolean flag to `list_config` that restores the strict validation crash behavior if the user explicitly requests it.
* **Updated `list_config` output:** The command now dynamically falls back to printing the raw dictionaries if the validated Pydantic models are unavailable.
* **Resilient Logging:** Added a fallback mechanism in `cmd_cli.py` to gracefully initialize the logger using raw dictionary values if Pydantic crashes during startup.

### Checklist
- [x] A changelog entry was added to `changelog.d/`

### DIFF
--- a/src/docbuild/cli/cmd_cli.py
+++ b/src/docbuild/cli/cmd_cli.py
@@ -90,14 +90,10 @@ def handle_validation_error(
 def load_app_config(
     ctx: click.Context,
     app_config: Path,
-    max_workers: str | None
+    max_workers: str | None,
+    skip_validation: bool = False
 ) -> None:
-    """Load and validate Application configuration.
-
-    :param ctx: The Click context object. The result will be added to ``ctx.obj.appconfig``.
-    :param app_config: The path to the application config file provided via CLI.
-    :param max_workers: The max_workers value from CLI options.
-    """
+    """Load and validate Application configuration."""
     context = ctx.obj
     result = handle_config(
         app_config,
@@ -110,18 +106,27 @@ def load_app_config(
         tuple[tuple[Path, ...] | None, dict[str, Any], bool], result
     )
 
+    # Always store the raw dict so `config list` has access to it
+    context.raw_appconfig = raw_appconfig
+
     if max_workers is not None:
         raw_appconfig["max_workers"] = max_workers
 
-    context.appconfig = AppConfig.from_dict(raw_appconfig)
+    try:
+        context.appconfig = AppConfig.from_dict(raw_appconfig)
+    except (ValueError, ValidationError) as e:
+        if skip_validation:
+            log.warning("Application config validation failed. Proceeding with raw data.")
+        else:
+            raise e
 
 
-def load_env_config(ctx: click.Context, env_config: Path) -> None:
-    """Load and validate Environment configuration.
-
-    :param ctx: The Click context object. The result will be added to ``ctx.obj.envconfig``.
-    :param env_config: The path to the environment config file provided via CLI.
-    """
+def load_env_config(
+    ctx: click.Context,
+    env_config: Path,
+    skip_validation: bool = False
+) -> None:
+    """Load and validate Environment configuration."""
     context = ctx.obj
     result = handle_config(
         env_config,
@@ -134,7 +139,18 @@ def load_env_config(ctx: click.Context, env_config: Path) -> None:
         tuple[tuple[Path, ...] | None, dict[str, Any], bool], result
     )
 
-    context.envconfig = EnvConfig.from_dict(raw_envconfig)
+    # Always store the raw dict so `config list` has access to it
+    context.raw_envconfig = raw_envconfig
+
+    try:
+        context.envconfig = EnvConfig.from_dict(raw_envconfig)
+    except (ValueError, ValidationError) as e:
+        if skip_validation:
+            log.warning("Environment config validation failed. Proceeding with raw data.")
+        else:
+            raise e
+
+
 
 @click.group(
     name=APP_NAME,
@@ -191,7 +207,7 @@ def load_env_config(ctx: click.Context, env_config: Path) -> None:
     ),
 )
 @click.pass_context
-def cli(
+def cli( # noqa: C901
     ctx: click.Context,
     verbose: int,
     dry_run: bool,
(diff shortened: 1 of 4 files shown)


tomschr (commented):
Thank you very much for all the work! 🎉

If you have no ENV config file in the repo, everything works as expected.

However, as long as there is one, it fails. I've tried it with an `env.production.toml` file and I get this traceback:

```
$ docbuild config list --flat
Environment config validation failed. Proceeding with raw data.
[...]
  File "/home/toms/repos/GH/opensuse/docbuild/src/docbuild/logging.py", line 214, in setup_logging
    handlers = build_handlers_from_config(config)
  File "/home/toms/repos/GH/opensuse/docbuild/src/docbuild/logging.py", line 163, in build_handlers_from_config
    handler = cls(**handler_args)
  File "/home/toms/.local/share/uv/python/cpython-3.14.6-linux-x86_64-gnu/lib/python3.14/logging/__init__.py", line 1219, in __init__
    StreamHandler.__init__(self, self._open())
                                 ~~~~~~~~~~^^
  File "/home/toms/.local/share/uv/python/cpython-3.14.6-linux-x86_64-gnu/lib/python3.14/logging/__init__.py", line 1248, in _open
    return open_func(self.baseFilename, self.mode,
                     encoding=self.encoding, errors=self.errors)
FileNotFoundError: [Errno 2] No such file or directory: '/home/toms/.local/state/docbuild/{server.name}/log/docbuild_2026-08-07_18-15-14.log'
```

It seems, the placeholders aren't replaced. Another problem could arise, when the ENV config uses a directory, that cannot be created or accessed.

I'm not sure how to solve this. Maybe we need to modify the `EnsureWriteableDirectory` at runtime. 🤔

sushant-suse:
> Thank you very much for all the work! 🎉
> 
> If you have no ENV config file in the repo, everything works as expected.
> 
> However, as long as there is one, it fails. I've tried it with an `env.production.toml` file and I get this traceback:
> 
> ```
> $ docbuild config list --flat
> Environment config validation failed. Proceeding with raw data.
> Traceback (most recent call last):
> [...]
>   File "/home/toms/repos/GH/opensuse/docbuild/src/docbuild/logging.py", line 214, in setup_logging
>     handlers = build_handlers_from_config(config)
>   File "/home/toms/repos/GH/opensuse/docbuild/src/docbuild/logging.py", line 163, in build_handlers_from_config
>     handler = cls(**handler_args)
>   File "/home/toms/.local/share/uv/python/cpython-3.14.6-linux-x86_64-gnu/lib/python3.14/logging/__init__.py", line 1219, in __init__
>     StreamHandler.__init__(self, self._open())
>                                  ~~~~~~~~~~^^
>   File "/home/toms/.local/share/uv/python/cpython-3.14.6-linux-x86_64-gnu/lib/python3.14/logging/__init__.py", line 1248, in _open
>     return open_func(self.baseFilename, self.mode,
>                      encoding=self.encoding, errors=self.errors)
> FileNotFoundError: [Errno 2] No such file or directory: '/home/toms/.local/state/docbuild/{server.name}/log/docbuild_2026-08-07_18-15-14.log'
> ```
> 
> It seems, the placeholders aren't replaced. Another problem could arise, when the ENV config uses a directory, that cannot be created or accessed.
> 
> I'm not sure how to solve this. Maybe we need to modify the `EnsureWriteableDirectory` at runtime. 🤔


Excellent catch Toms. I just pushed a fix. Now, whenever the CLI is in "skip validation" mode (where we are just dumping the raw dictionary to the terminal), it intercepts the logging setup and forces a safe, console-only logging configuration with no file handlers. This completely eliminates the risk of placeholder or directory permission crashes while debugging. Let me know if it runs smoothly for you now.

src/docbuild/cli/cmd_config/list.py
```diff
@@ -25,19 +25,32 @@ def print_section(title: str, data: dict[str, Any], prefix: str, flat: bool, col
...
+    if env or show_all:
+        if context.envconfig:
+            env_data = context.envconfig.model_dump(mode="json")
+        elif context.raw_envconfig:
+            env_data = context.raw_envconfig
+        else:
+            env_data = {}
+
+        if env_data:
+            print_section("Environment Configuration", env_data, "env", flat, "yellow")
```
tomschr: Hmn... could be considered a code smell. 🤔 I've tried to avoid code duplication, but my solution isn't really better:


```suggestion
    # We group the varying parameters into a list of configurations to process
    sections_to_check = [
        (
            app or show_all,
            context.appconfig,
            context.raw_appconfig,
            "Application Configuration",
            "app",
            "cyan",
        ),
        (
            env or show_all,
            context.envconfig,
            context.raw_envconfig,
            "Environment Configuration",
            "env",
            "yellow",
        ),
    ]

    for should_show, config, raw_config, title, key, color in sections_to_check:
        if not should_show:
            continue

        # Extract data if config exists, fallback to raw_config, or use empty dict
        data = config.model_dump(mode="json") if config else (raw_config or {})

        if data:
            print_section(title, data, key, flat, color)
```

I leave it up to you. 😄

src/docbuild/cli/cmd_cli.py
```diff
@@ -235,27 +252,52 @@ def cli(
     current_model: type[BaseModel] = AppConfig
     current_files: Sequence[Path] | None = None
 
+    # Determine if we should skip validation
+    is_config_list = ctx.invoked_subcommand == "config" and "list" in sys.argv
```
tomschr: For Python the operator order is clear, but I accidentely read `"config" and "list"` which doesn't make sense. 😉

A small nitpick, maybe group that to make it visually obvious:

```suggestion
    is_config_list = (ctx.invoked_subcommand == "config") and ("list" in sys.argv)
```

tomschr (commented):
Very good, Sushant! It works for me! :tada: 

I just have two nitpicks. The first could be considered as a matter of taste (you can ignore that). It's just a crazy idea and don't fly. The second is more about code aesthetics. It's up to you if you want to include them.

Ready to be merged! :+1:

src/docbuild/cli/cmd_cli.py
sushant-suse: It make sense. I have updated it. Thanks.

src/docbuild/cli/cmd_config/list.py
sushant-suse: I actually really like your refactor. The ternary operator fallback is much cleaner than the nested if/else blocks I had, and keeping things DRY is always a win. I've applied your suggestion and pushed the update. Thanks 😄

tomschr (approved):

PR #431, opened 2026-08-10
Refactor Doctype.xpath and list_all_deliverables

tomschr:
Collect all XPath related parts in Doctype and NOT spread it to other functions.

Changes:
* Function `list_all_deliverables` Refactor and avoid manually creating XPath expression. Better rely on `Doctype.xpath()`

* Class `Doctype`
  - Introduce lifecycle_xpath_segment() and locale_xpath_segment. The two returns the respective XPath expression for the lifecycle and the locale part
  - Refactor .xpath() and call the respective *_xpath_segment() methods internally
  - Add an optional absolute argument (default = False) to add "//" in front of the XPath

### DIFF
--- a/src/docbuild/models/doctype.py
+++ b/src/docbuild/models/doctype.py
@@ -35,6 +35,7 @@ class Doctype(BaseModel):
     >>> doctype = Doctype.from_str("sles/15-SP6@supported/en-us,de-de")
     >>> doctype.product
     <Product.sles: 'sles'>
+    <Product.sles: 'SUSE Linux Enterprise Server'>
     >>> doctype.docset
     ['15-SP6']
     >>> doctype.lifecycle.name
@@ -78,6 +79,7 @@ class Doctype(BaseModel):
         default=[LanguageCode(language="en-us")],
         description=(
             "The natural language containing language and country. "
+            "Values can be combined using commas. "
             "After validation, langs are sorted"
         ),
         examples=["en-us", "de-de"],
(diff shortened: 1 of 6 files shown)


src/docbuild/models/doctype.py
```diff
@@ -281,11 +261,47 @@ def product_xpath_segment(self: Self) -> str:
...
+            return f"[{lifecycle}]"
+        return ""
+
+    def locale_xpath_segment(self: Self) -> str:
+        """Return the XPath segment for the language node.
+
+        Example: "resources/locale[@lang='en-us']" or "resources/locale"
+        """
+        language = "locale"
+        if self.langs and "*" not in self.langs:
```
sushant-suse: Hi Toms, since `self.langs` is a list of `LanguageCode` objects (not raw strings), checking `if "*" not in self.langs:` might fail or behave unpredictably depending on how Pydantic implements `__eq__` for that model.

In your previous PR, you used a safer generator approach for this exact check: `any(lang.language == "*" for lang in self.langs)`.

I know you just moved this code from the old `.xpath()` method so it's a pre-existing quirk, but it might be worth updating it here to match the safer approach:
```python
has_wildcard = any(lang.language == "*" for lang in self.langs)
if self.langs and not has_wildcard:
```

sushant-suse (approved):
Hi Toms, this refactor is incredibly clean. I just left one small note about how we check for language wildcards, but otherwise, this is an easy approval. Thanks

src/docbuild/models/doctype.py
tomschr: Great catch! :heart_eyes: You are absolutely right. I changed it accordingly and amended a test case in `test_xpath_in_doctype()`. All committed in 09d6859c.

Issue #370, opened 2026-07-14
Restructure obsolete sections in env config


tomschr:
### Situation

The env configs still have a `[server]` section in the TOML file. However, we are not a server anymore. This was from the old config.

### Use Case

* Reduce confusion for misleading terms/structures

### Possible Implementation

* Remove the `server` section.
* Remove `host` and `port` key. We don't need that anymore.
* Keep `name`, `role`, and `enable_mail`.
* For the `config section, maybe find a better name (it's _all_ config).
* Not sure about the `xslt-params` section. Should we split it into a common key that is for *all* target formats, and specific for HTML and FO/PDF? With the latter part, we could better distinguish between parameters for different target formats. => Maybe this is something for a different issue/PR?

Check list:

* [x] Going through all tests and adjust the names.
* [x] Going through all code and adjust the names.
* [x] Adjust the env config model(s) in `src/docbuild/models/config/env.py`.
* [x] Adjust the example file in `etc/docbuild/env.example.toml`.
* [x] Perhaps adjust documentation.

PR #433, opened 2026-08-11
feat: allow setting env config via CLI with -C/--set-env

### LINKED ISSUE #410: Allow to set env config through the command line option
### Situation

We have the default settings in `DEFAULT_ENV_CONFIG` and we can pass env config file(s). However, if we only want to change one single key it's a bit cumbersome to just add it as a file.

### Use Case

As a user, I want to change the default env config options on the CLI to that I don't have to create a config file.

### Possible Implementation

* [x] Add a global command line option `--env-config`/`-C` with the syntax `key=value`. For nested configs, use the dot notation like `paths.config_dir`.
* [x] Unknown keys aren't allowed and lead to a `ValueError` that is shown to the user.
* [x] The value of `key=value` needs to be potentially masked or converted depending on the key. I guess, Pydantic could do that?
* [x] Priority from low to high is `DEFAULT_ENV_CONFIG` < system config < user config < current directory < CLI option. Options from the right overwrite options from the left.
* [x] All config objects are merged.

sushant-suse:
Closes #410

**What was done:**

* **Added `-C`/`--set-env` option:** Introduced a new multiple-value option to the root `cli()` command, allowing users to pass `key=value` configuration overrides.
* **Implemented dotted notation parsing:** Updated `load_env_config` to split dotted keys (e.g., `paths.tmp.log_dir=/tmp`) and deeply merge them into the raw environment configuration dictionary.
* **Utilized Pydantic for validation:** Injected the CLI overrides *before* `EnvConfig.from_dict()` is called. This safely handles automatic type coercion (e.g., `"true"` to `True`) and automatically rejects unknown keys because of Pydantic's strict model settings.
* **Cleaned up CLI setup:** Removed an accidental duplicate call to `load_env_config` in `cmd_cli.py`.
* **Updated Test Suite:** Updated `test_cmd_cli.py` to satisfy the new `load_env_config` signature and added `test_load_env_config_applies_cli_overrides` to verify correct override parsing and `click.BadParameter` handling.

### Checklist

* [x] Code is properly formatted (`uvx ruff check --fix`)

### DIFF
--- a/docs/source/user/config.rst
+++ b/docs/source/user/config.rst
@@ -206,6 +206,6 @@ Use ``key=value`` syntax, and use dot notation for nested options:
 
 .. code-block:: shell-session
 
-    $ docbuild -C server.host=127.0.0.1 -C server.enable_mail=true config list --env
+    $ docbuild -C "config.canonical_url_domain=https://doc.example.net"  -C paths.root_config_dir=/etc/docbuild config list --env
 
 Options passed via ``-C`` or ``--set-env`` have the highest priority and will strictly overwrite the corresponding settings loaded from your default, system, or local configuration files.
--- a/docs/source/user/config.rst
+++ b/docs/source/user/config.rst
@@ -193,3 +193,19 @@ between configurations without needing to remember the exact command syntax.
    $ alias docbuild-prod='docbuild --env-config env.production.toml'
    $ alias docbuild-test='docbuild --env-config env.testing.toml'
    $ alias docbuild-dev='docbuild --env-config env.devel.toml'
+
+
+.. _config-overwriting-cli:
+
+Overwriting Config Options on the Command Line
+----------------------------------------------
+
+You can temporarily override environment configuration options directly from the command line using the ``-C`` or ``--set-env`` option. This avoids having to create or modify TOML files for quick adjustments.
+
+Use ``key=value`` syntax, and use dot notation for nested options:
+
+.. code-block:: shell-session
+
+    $ docbuild -C server.host=127.0.0.1 -C server.enable_mail=true config list --env
+
+Options passed via ``-C`` or ``--set-env`` have the highest priority and will strictly overwrite the corresponding settings loaded from your default, system, or local configuration files.
--- a/tests/cli/test_cmd_cli.py
+++ b/tests/cli/test_cmd_cli.py
@@ -313,5 +313,28 @@ def test_load_config_helpers_populate_context(fake_handle_config, mock_config_mo
     assert mock_ctx.obj.appconfig is not None
 
     # Test Env Loader
-    load_env_config(mock_ctx, Path("env.toml"))
+    load_env_config(mock_ctx, Path("env.toml"), env_overrides=())
     assert mock_ctx.obj.envconfig is not None
+
+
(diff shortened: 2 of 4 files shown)


tests/cli/test_cmd_cli.py
```diff
@@ -313,5 +313,28 @@ def test_load_config_helpers_populate_context(fake_handle_config, mock_config_mo
     assert mock_ctx.obj.appconfig is not None
 
     # Test Env Loader
-    load_env_config(mock_ctx, Path("env.toml"))
+    load_env_config(mock_ctx, Path("env.toml"), env_overrides=())
     assert mock_ctx.obj.envconfig is not None
+
+
+def test_load_env_config_applies_cli_overrides(fake_handle_config, mock_config_models):
```
tomschr, line 320: There is no tests when the user just provides a key (without a `=` sign).

tomschr (changes requested):
That's a great addition! Well done Sushant! 👍 🎉 

Apart from the suggestion below, I think it would be beneficial to mention this new option in the documentation. Maybe add a new section in the User Guide? Maybe use the title `Overwriting config options on the commandline` or similar? It doesn't have to be long.

I just saw that the User Guide's section should be a bit clearer in the distinction between ENV/App and Portal config. But that's for a different issue. 😉

tests/cli/test_cmd_cli.py
sushant-suse: Adding it. Thanks

sushant-suse:
> That's a great addition! Well done Sushant! 👍 🎉
> 
> Apart from the suggestion below, I think it would be beneficial to mention this new option in the documentation. Maybe add a new section in the User Guide? Maybe use the title `Overwriting config options on the commandline` or similar? It doesn't have to be long.
> 
> I just saw that the User Guide's section should be a bit clearer in the distinction between ENV/App and Portal config. But that's for a different issue. 😉

Thanks for the feedback Toms

I've updated the PR with:
1. A dedicated CLI test (`test_load_env_config_missing_equals_sign_fails`) verifying that passing `-C key` without an `=` sign correctly raises an error.
2. Added a new section **'Overwriting config options on the commandline'** to the User Guide explaining `-C`/`--set-env` with dot notation examples.

It is ready for another look. Thanks.

docs/source/user/config.rst
```diff
@@ -193,3 +193,19 @@ between configurations without needing to remember the exact command syntax.
...
+Overwriting Config Options on the Command Line
+----------------------------------------------
+
+You can temporarily override environment configuration options directly from the command line using the ``-C`` or ``--set-env`` option. This avoids having to create or modify TOML files for quick adjustments.
+
+Use ``key=value`` syntax, and use dot notation for nested options:
+
+.. code-block:: shell-session
+
+    $ docbuild -C server.host=127.0.0.1 -C server.enable_mail=true config list --env
```
tomschr: I would use different keys as these will be subject to removal in #370.


```suggestion
    $ docbuild -C "config.canonical_url_domain=https://doc.example.net"  -C paths.root_config_dir=/etc/docbuild config list --env
```

tomschr (changes requested):
Great job, Sushant! 👍

Just a minor suggestion for the documentation. Once this is done, merge it and we are done! 👍 🎉

Issue #448, opened 2026-08-14
Product shows full product names instead of acronyms


tomschr:
### Project Version

0.22.0

### Python Version

3.12, 3.13, 3.14

### Operating System

Generic

### Steps to Reproduce

Call this:

```shell
$ docbuild -vv metadata 'sles-ha/16.0/'
```

### Actual Behavior

The above command creates this output:

```
ERROR: Invalid doctype string 'sles-ha/16.0/': 'sles-ha' is not a valid Product. Allowed values are: '*', 'Appliance building', 'Cloud Native', 'Compliance Documentation', 'Container Documentation', 'SUSE Multi-Linux Support', 'SUSE Release Notes', 'SUSE Best Practices', 'SUSE Enterprise Storage', 'SUSE Linux Enterprise Desktop', 'SUSE Linux Enterprise High Availability', 'SUSE Linux Enterprise High-Performance Computing', 'SUSE Linux Micro', 'SUSE Linux Enterprise in Public Clouds', 'SUSE Linux Enterprise Real Time', 'SUSE Linux Enterprise Server for SAP applications', 'SUSE Linux Enterprise Server', 'SUSE Linux Enterprise Virtual Machine Driver Pack', 'SUSE Smart Docs', 'SUSE Linux Enterprise Subscription Management Tool', 'SUSE OpenStack Cloud', 'SUSE Documentation Style Guide', 'Subscription Management', 'SUSE Multi-Linux Manager for Retail', 'SUSE Multi-Linux Manager', 'SUSE AI Factory', 'SUSE AI', 'SUSE CaaS Platform', 'SUSE Cloud Application Platform', 'SUSE Distribution Migration System', 'SUSE Edge', 'SUSE Telco Cloud', 'Technical Reference Documentation'
Aborted!
```

### Expected Behavior

The products should be NOT written with full product names. Instead the acronyms should be displayed:

```
[...]

PR #449, opened 2026-08-14
List product acronyms in invalid-value errors

tomschr:
Adjusted invalid ``Product`` error messages during doctype parsing to list allowed product acronyms (for example ``sle-ha`` and ``sles``) instead of full product names.

Changes:

* Adjust `src/docbuild/models/product.py` to use the acronyms instead the full product names
* Adjust `tests/models/test_product.py`


sushant-suse (approved):

tomschr:
Fixes #448

Issue #447, opened 2026-08-14
Idea: Implement dependency-aware caching for DAPS metadata extraction


tomschr:
### Situation

Currently, extracting metadata from documentation deliverables using `daps metadata` is a time-consuming process. When managing multiple deliverables across various projects and languages, the pipeline must:

1. Update Git repositories.
1. Create temporary working directories.
1. Execute daps metadata on each deliverable.

Because documentation source files are highly modular, a single deliverable relies on many referenced XML, image, and entity files. Currently, `daps metadata` is re-executed on every run—even when none of the underlyinsg source files have changed—leading to significant, unnecessary build and processing overhead.

### Use Case / User Story

As a developer, I want to check whether a deliverable's source files have changed before running `daps metadata`, so that I can reuse cached metadata and skip expensive daps executions for unchanged deliverables.

### Possible Implementation

```
  ┌──────────────────────────────┐
  │ Check if cached record exists│
  └──────────────┬───────────────┘
                 │
   ┌─────────────┴─────────────┐
   ▼                           ▼
[ Cache Miss ]               [ Cache Hit ]
   │                           │
   │                           ▼
   │             Read recorded file dependencies 
   │             from cache file & hash them.
   │                           │
   │             ┌─────────────┴─────────────┐
   │             ▼                           ▼
   │         [ Hash Match ]              [ Hash Mismatch ]
   │             │                           │
   │             ▼                           │
   │      Return Cached Metadata             │
   │      (Skipped DAPS run!)                │
   │                                         │
   └────────────────────┬────────────────────┘
                        │
                        ▼
   Execute `daps metadata` & `daps list-src`
   Save metadata + new dependencies + new hash
```

1. **First Execution (Cache Miss):** You run `daps metadata` and retrieve the source files using `daps list-srcfiles`. Compute the fingerprint for all those XML & ADoc files and write (metadata, dependencies, hash) to your cache directory.
2. **Subsequent Executions:** Look up the cached dependency list for the target DC file. Hash those specific files on disk/Git. If the hash matches the cached hash, return the cached metadata immediately without running daps.
3. **Dependency Change (Cache Invalidation):** If an XML file or include was modified, the hash will mismatch. The code will re-run daps, parse the new metadata and updated dependency list, and refresh the cache.
4. **Potential new subcommand:** Perhaps it would be helpful to have a `docbuild cache list|update|clean` 

[...]

Issue #154, opened 2026-02-03
Life display/progress view when building deliverables


tomschr:
### Situation

When building deliverables in verbose output, the output can be overwhelming. Additionally, when something goes wrong it could be already gone out of sight.

### Use Case

It would be helpful to structure the output in a way that it allows to:

* see which deliverable is currently being built,
* see any errors at the same time.

This would be an _additional_ view for the traditional output with logging.

### Possible Implementation

We could use the `rich` library (already used as a dependency). I'm currently not 100% sure how this could look like. There are many ways to group that. This needs further discussion.

As a working draft, the output screen could look like this:

```
[▌▌▌▌    ]  sles/16.0/en-us
[▌▌▌     ]  sle-ha/16.0/en-us
[▌▌▌▌▌▌▌▌]  suse-ai/1.0/en-us       ✅

--------------
sles/16.0/en-us:DC-foo-bar: ERROR wrong
```

The first pane contains a progress bar that shows all requested doctypes (=product & docset & lang). When everything was successfully built, it gets a checkmark (✅).

The second pane contains error messages. It can be validation errors, JSON problems, or any other issue that went wrong. Maybe even warnings. Everything that is needed to judge if it's a serious issue.

### Further information

https://rich.readthedocs.io/en/stable/live.html

PR #450, opened 2026-08-14
refactor(metadata): implement aiostream pipeline for async task orchestration

### LINKED ISSUE #325: Research: Async "Pipelines"
### Situation

This is more of a research project than a real implementation. But if there is a proof-of-concept, that would be wonderful!

I have the feeling that stacking async functions together gets more and more diffcult. As docbuild uses different tasks that build on each other, it may be useful to implement a kind of "pipeline" approach, similar to the Unix pipes.

### Use Case

Helps to build more sophisticated and flexible workflows.

### Possible Tools

There are several external libraries which we could use, but they all have their pros and cons (or tradeoffs):

* [aiostream](https://github.com/vxgmichel/aiostream), [docs](http://aiostream.readthedocs.io/)

  Best for:
  
  * ✔️ async function chaining
  * ✔️ composable pipelines
  * ✔️ elegant syntax
  * ✔️ lightweight

* [anyio](https://github.com/agronholm/anyio), [docs](https://anyio.readthedocs.io)

  Best for:

  * ✔️ modern structured concurrency
  * ✔️ cancellation correctness
  * ✔️ async worker systems
 
* [trio](https://github.com/python-trio/trio), [docs](https://trio.readthedocs.io/)

  Best for:

  * ✔️ a simpler, cleaner API
  * ✔️ Predictable and safe cancellation
  * ✔️ Structural concurrency (no orphan tasks)

The latter is not necessarily a pipeline tool per se, but a complete _replacement_ of the standard `asyncio` lib. Nevertheless, I thought it could fit into this issue. 

### Research points

* Learning curve
* Creating sub-tasks
* Cancellation of tasks
* Exception handling

Comparisons

* Trio is a full async _runtime_ built from scratch around structured concurrency.
* AnyIO is a compatibility layer that sits on top of either asyncio or Trio. It allows you to write code once using a Trio-like structured API, and run it anywhere.
* aiostream is a utility library specifically for asynchronous iteration (like `itertools` or reactive streams for async generators). It works beautifully with asyncio, but has some friction with structured runtimes.

sushant-suse:
Closes #325 

**What was done:**

* **Researched Async Pipelines:** Evaluated `AnyIO`, `Trio`, and `aiostream` for pipeline orchestration as requested in the issue. 
* **Implemented Proof-of-Concept:** Selected `aiostream` as it perfectly supports the requested Unix-style `pipe1 | pipe2 | pipe3` chaining syntax using overloaded bitwise OR operators.
* **Refactored `runner.py`:** Replaced the legacy manual `asyncio.Semaphore` and explicit task collection loops (`run_tasks_fail_fast`, `run_tasks_collect_all`) with a clean `aiostream` pipeline.
* **Maintained Concurrency Safety:** The new pipeline natively respects `max_workers` limits and automatically handles safe cancellation of pending tasks when short-circuiting on failures (`exitfirst=True`).
* **Cleaned up tests:** Removed obsolete unit tests for the deleted manual task handlers while ensuring top-level pipeline coverage remains intact.
* **Added Documentation:** Created `async-pipeline.rst` in the Developer Guide under Concepts, featuring a Mermaid flowchart that visualizes the new pipeline architecture.

### Checklist

* [x] Documentation has been updated

### DIFF
--- a/docs/source/developer/async-pipeline.rst
+++ b/docs/source/developer/async-pipeline.rst
@@ -25,9 +25,44 @@ The following flowchart illustrates how deliverables stream through the metadata
        %% Define the output
        P2 -->|success, deliverable| Output[/Failure Collection & Early Exit/]
 
-       %% Styling
-       style Async Pipeline fill:#f9f9f9,stroke:#333,stroke-width:2px
-       style Iter fill:#e1f5fe,stroke:#0288d1
-       style P1 fill:#fff3e0,stroke:#f57c00
-       style P2 fill:#e8f5e9,stroke:#388e3c
-       style Output fill:#fce4ec,stroke:#c2185b
+
+Implementation & Exception Handling
+-----------------------------------
+
+Because ``aiostream`` expects a mapping function to process items cleanly, we wrap our core execution functions to catch exceptions and return structured status tuples (e.g., ``(success_boolean, deliverable)``). This prevents a single failed deliverable from ungracefully crashing the entire pipeline.
+
+Here is a simplified example of how we interact with ``aiostream`` and handle errors:
+
+.. code-block:: python
+
+    from aiostream import stream, pipe
+
+    async def my_task_wrapper(deliverable, *args: object) -> tuple[bool, object]:
+        """Wrapper to catch exceptions safely so the pipeline continues."""
+        try:
+            # Attempt the actual heavy-lifting task
+            await process_deliverable(deliverable)
+            return True, deliverable
+        except Exception as e:
+            # Log the error and return a failure state
+            log.error(f"Task failed: {e}")
+            return False, deliverable
+
+    async def run_pipeline(deliverables):
+        # 1. Create the stream
+        pipeline = stream.iterate(deliverables) | pipe.map(
+            my_task_wrapper, task_limit=8
+        )
+
+        failed_items = []
+
+        # 2. Consume the pipeline
+        async with pipeline.stream() as streamer:
+            async for success, deliverable in streamer:
+                if not success:
+                    failed_items.append(deliverable)
+                    # To "fail fast", we can simply break the loop. 
+                    # aiostream automatically safely cancels all pending tasks!
+                    # break 
+
+        return failed_items
--- a/docs/source/developer/async-pipeline.rst
+++ b/docs/source/developer/async-pipeline.rst
@@ -1,0 +1,33 @@
+Async Pipeline Architecture
+===========================
+
+DocBuild uses a pipeline architecture for orchestrating complex, concurrent tasks (like fetching repositories, extracting metadata, and building deliverables).
+
+To achieve this, we use the ``aiostream`` library. It allows us to chain asynchronous generators using a Unix-pipe-like syntax (``|``), automatically handling worker limits, graceful cancellation, and task fan-out.
+
+Pipeline Flow
+-------------
+
+The following flowchart illustrates how deliverables stream through the metadata extraction pipeline concurrently:
+
+.. mermaid::
+
+   graph TD
+       %% Define the input stream
+       Input[/List of Deliverables/] --> Iter[stream.iterate]
+
+       %% Define the pipeline steps
+       subgraph Async Pipeline [aiostream Unix-like Pipeline]
+           Iter -->|deliverable| P1(pipe.map: update_repositories)
+           P1 -->|repo_dir| P2(pipe.map: process_deliverable_wrapper)
+       end
+
+       %% Define the output
+       P2 -->|success, deliverable| Output[/Failure Collection & Early Exit/]
+
+       %% Styling
+       style Async Pipeline fill:#f9f9f9,stroke:#333,stroke-width:2px
+       style Iter fill:#e1f5fe,stroke:#0288d1
+       style P1 fill:#fff3e0,stroke:#f57c00
+       style P2 fill:#e8f5e9,stroke:#388e3c
+       style Output fill:#fce4ec,stroke:#c2185b
--- a/src/docbuild/tasks/metadata/runner.py
+++ b/src/docbuild/tasks/metadata/runner.py
@@ -69,7 +69,7 @@ async def process_doctype(
     )
 
     # Sort deliverables alphabetically for predictable processing order
-    deliverables.sort(key=lambda d: d.full_id)
+    deliverables.sort()
 
     if skip_repo_update:
         log.info("Skipping repository %s updates as requested.", repo_dir)
--- a/src/docbuild/tasks/metadata/runner.py
+++ b/src/docbuild/tasks/metadata/runner.py
@@ -5,7 +5,8 @@ from collections.abc import Sequence
 import logging
 from pathlib import Path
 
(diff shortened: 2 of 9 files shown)


docs/source/developer/async-pipeline.rst
```diff
@@ -0,0 +1,33 @@
...
+
+       %% Define the output
+       P2 -->|success, deliverable| Output[/Failure Collection & Early Exit/]
+
+       %% Styling
+       style Async Pipeline fill:#f9f9f9,stroke:#333,stroke-width:2px
+       style Iter fill:#e1f5fe,stroke:#0288d1
+       style P1 fill:#fff3e0,stroke:#f57c00
+       style P2 fill:#e8f5e9,stroke:#388e3c
+       style Output fill:#fce4ec,stroke:#c2185b
```
tomschr: The colors doesn't seem to work on a dark theme. They are very faint:

Image

tomschr: This is a great start! Perhaps what's missing is how we interact with aiostream and how we catch exceptions. Maybe a short example would also be nice?

tomschr (commented):
Sushant, this is brilliant! 👍 🚀 😍 I've tested it with different doctypes on SLES to get the elapsed time:

* Doctype `sles/16.0/`
  Elapsed time 34.7s

* Doctype `sles/15-SP7/`
  Elapsed time 145s

* Doctype `sles/*/`
  Elapsed time 2344.03s => 39min (!)

Note: For SLES 15-SP7, docbuild had to clone the repo first whereas for 16.0 it was mostly there. Same for `sles/*/`.

What would be cool if the order of the deliverables is somehow predictable. Maybe this is not that possible (or desired), but with a predictable order you can kind of follow and can estimate what's still left.

~Perhaps this is for a different PR (and I should open a new issue for that), but when I try to build the metadata for a large set (like `sles/*/`), I'd like to know where I am and what's left. A kind of progress bar.~ Also related to is issue #154.

I think this is outside of THIS specific PR. Not sure if it would be feasible to kind of "prepare" for that use case.

docs/source/developer/async-pipeline.rst
sushant-suse: Agree Toms. I added it.

sushant-suse:
> Sushant, this is brilliant! 👍 🚀 😍 I've tested it with different doctypes on SLES to get the elapsed time:
> 
> * Doctype `sles/16.0/`
>   Elapsed time 34.7s
> * Doctype `sles/15-SP7/`
>   Elapsed time 145s
> * Doctype `sles/*/`
>   Elapsed time 2344.03s => 39min (!)
> 
> Note: For SLES 15-SP7, docbuild had to clone the repo first whereas for 16.0 it was mostly there. Same for `sles/*/`.
> 
> What would be cool if the order of the deliverables is somehow predictable. Maybe this is not that possible (or desired), but with a predictable order you can kind of follow and can estimate what's still left.
> 
> ~Perhaps this is for a different PR (and I should open a new issue for that), but when I try to build the metadata for a large set (like `sles/*/`), I'd like to know where I am and what's left. A kind of progress bar.~ Also related to is issue #154.
> 
> I think this is outside of THIS specific PR. Not sure if it would be feasible to kind of "prepare" for that use case.

Thanks for doing those benchmarks. 39 minutes for the entirety of `sles/*/` is actually really impressive given how heavy `daps metadata` is. It proves exactly why we need the dependency caching from #447 next.

Regarding your thoughts on the progress bar and ordering, this PR actually perfectly prepares us for that:

1. **Predictable Ordering:** We can sort the initial `deliverables` list alphabetically before feeding it into the stream. `aiostream` also has an `ordered=True` argument for its pipes if we ever strictly need the output to match the input order.
2. **Progress Bar (#154):** Because `aiostream` yields items dynamically as they finish in the `async for success, deliverable in streamer:` loop, we can trivially wrap that loop in a `rich.progress` bar in the future. Every yield updates the bar!

**Documentation updates**:
I have pushed an update to the PR that addresses your review:

1. I removed the hardcoded HTML colors in the Mermaid diagram so it should now dynamically respect GitHub's and Sphinx's native light/dark themes.
2. I added an **"Implementation & Exception Handling"** section to the `.rst` file with a Python code block explaining the wrapper pattern, exception catching, and early exit behavior.

Thanks

docs/source/developer/async-pipeline.rst
tomschr: Works now! Looks much better! Thank you! :+1:

tomschr:
That's great! :tada: Thank you!

1. Predictable Ordering. I think we should do that as this would simplify the progress bar handling, doesn't it?
2. Progress Bar. Perfect!

Do we need to adjust the code for point 1?

sushant-suse:
> That's great! 🎉 Thank you!
> 
> 1. Predictable Ordering. I think we should do that as this would simplify the progress bar handling, doesn't it?
> 2. Progress Bar. Perfect!
> 
> Do we need to adjust the code for point 1?

To answer your questions:

1. The progress bar logic technically doesn't strictly *need* ordered output (since it just increments a `count / total` as things finish), but you are 100% right that from a UX perspective, predictable ordering makes the terminal output vastly easier for a developer to follow and estimate!
2. Yes, we should adjust the code for that right now. I just pushed a quick update to the PR!

I made two small additions to `runner.py`:

* `deliverables.sort(key=lambda d: d.full_id)`: Guarantees the list is fed into the pipeline in alphabetical order.
* `pipe.map(..., ordered=True)`: Instructs `aiostream` to yield the results back to us in that exact same sorted order, even if a later deliverable finishes extracting faster than an earlier one.

Also updated the test accodingly.

This completely preps the pipeline for #154's progress bar.
Thanks.

src/docbuild/tasks/metadata/runner.py
```diff
@@ -129,18 +68,21 @@ async def process_doctype(
         get_deliverable_from_doctype, root, doctype
     )
 
+    # Sort deliverables alphabetically for predictable processing order
+    deliverables.sort(key=lambda d: d.full_id)
+
```
tomschr, line 73: You do the sorting with this code:

```python
deliverables.sort(key=lambda d: d.full_id)
```

That's possible and works. :+1: That kind of reminds me that we don't support sorting for the `Deliverable` class itself. 🤔 What if we provide that like this:

```python
from functools import cached_property, total_ordering
# ...

@total_ordering
@dataclass
class Deliverable:
    # ...
    def __eq__(self, other: object) -> bool:
        """Return True if the other deliverable has the same full ID."""
        if not isinstance(other, Deliverable):
            return NotImplemented
        return self.full_id == other.full_id

    def __lt__(self, other: object) -> bool:
        """Compare two Deliverable objects for sorting."""
        if not isinstance(other, Deliverable):
            return NotImplemented
        return self.full_id < other.full_id
```

Maybe there are better ways to do it. That would serve as a kind of "default sorting" and we would not have to add it all the time for `list.sort`.

What do you think?

tomschr (approved):
Great work! 😍 Lovin' it! 👍

(Squash-)Merge it, please!

Issue #12, opened 2025-07-04
Build deliverables


tomschr:
### Situation

We want to build deliverables from DocBook or ADoc sources to be published on d.s.c.

### Use Case

As a user, when I request a build of a specific project or the whole set, it performs several steps to create the output formats with the index pages in the target directory.

### Possible Implementation

The script perform several steps:

1. Read the XML Docserv config and stitch them together. Maybe store the result in a cache directory.
2. Determine the set of deliverables.
3. Iterate over it in a concurrent way.
4. Determine the repository of the deliverable. If there is no repo inside the `permanent-full` directory, clone it.
5. Clone a temporary repo from the `permanent-full` directory.
6. Create a temporary directory as the result directory, but use the temporary repo as a source.
7. If the build was successful, create the metadata. This could also be done prior to building the target formats.
8. Collect the metadata from the deliverable and other resources.
9. Create an intermediate JSON structure and pass it to Gayathri's Jinja templates.
10. Store the result in the temporary directory.
11. If everything was successful, sync it to the target directory ready for the final publishing step.

Issue #452, opened 2026-08-17
Implement rsync capabilities


tomschr:
### Situation

For `docbuild build` we need a way to synchronize directories. Usually we want to sync from a temporary directory into the target directory.

### Use Case / User Story

As a developer, I want an async `rsync` function so I can sync different directories easily.

### Possible Implementation

* [x] The async function requires at least the source and target paths as arguments. Potentially, we also want to pass optional arguments for rsync.
* [x] Decide if we want to add a config option to allow customizing the rsync call (recommended)
* [x] Include `rsync` tool into the checklist for `docbuild doctor`.
* [x] Ensure that rsync's trailing slash is covered (without slash = sync the directory and its content, with slash = sync only the content)
* [x] Consider introducing a `commands` section to collect dependent commands in the ENV config, e.g. `commands.rsync`. Could also be used to move `build.daps.command` to `commands.daps`.

As a starting point, here is one possible implementation:

```python
import asyncio
import os
import subprocess
from dataclasses import dataclass, field, fields
from pathlib import Path


@dataclass(frozen=True, slots=True)
class RsyncOptions:
    """Configuration options for rsync execution."""
    
    # Store the CLI flag directly in the field's metadata
    archive: bool = field(default=True, metadata={"flag": "-a"})
    compress: bool = field(default=False, metadata={"flag": "-z"})
    delete: bool = field(default=False, metadata={"flag": "--delete"})
    dry_run: bool = field(default=False, metadata={"flag": "--dry-run"})
    verbose: bool = field(default=False, metadata={"flag": "-v"})
    partial: bool = field(default=False, metadata={"flag": "--partial"})
    
    exclude: list[str] | tuple[str, ...] = field(
        default_factory=list, metadata={"flag": "--exclude"}
    )
    
    # Acts as an escape hatch. Notice it has no "flag" in metadata.
    extra_args: list[str] = field(default_factory=list)
    
    # Acts as an escape hatch for the 100+ rsync options not explicitly modeled here.
    # Users can pass arbitrary flags (e.g., ["--exclude=*.tmp", "--bwlimit=1000"]).
    extra_args: list[str] = field(default_factory=list)

    def to_args(self) -> list[str]:
        """Converts the configured options into a list of command-line arguments.
        
        :return: A list of string arguments formatted for the rsync command.
        """
        args: list[str] = []

        for f in fields(self):
            # Retrieve the flag string from metadata (returns None if not present)
            flag = f.metadata.get("flag")
[...]

PR #456, opened 2026-08-19
feat(build): implement async daps builder pipeline

sushant-suse:
Addresses #12

**What was done:**

* **Created Build CLI:** Implemented the `docbuild build` command using `click`, accepting doctype arguments and a `--skip-repo-update` flag to bypass git fetches during local testing.
* **Implemented Async Build Pipeline:** Leveraged our new `aiostream` architecture to concurrently process deliverables and iterate over their configured formats.
* **Wrapped Subprocess Execution:** Created robust async wrappers to execute the external `daps html` and `daps pdf` commands via `docbuild.utils.shell.run_command`, handling standard output, error codes, and exceptions.
* **Integrated Repository Management:** Hooked into the existing `update_repositories` logic to ensure required git repositories are cloned or updated before the build triggers.
* **Added Comprehensive Tests:** Wrote isolated unit tests utilizing `AsyncMock` to verify pipeline orchestration, formatting fallbacks, and execution failures, achieving 100% coverage for the new module.

@tomschr, this PR successfully implements the core concurrency engine to call DAPS (covering Steps 1-4 of the original issue description). Before moving on, I wanted to check how you'd like to handle the remaining steps (5-11):

1. **Output Dirs:** Should we adjust the DAPS command to output to a specific temporary directory (Step 6)?
2. **Chaining:** Should the `build` command automatically trigger the `metadata` task upon a successful build (Steps 7-10)?
3. **Sync:** Where should the final sync/publish step live (Step 11)?

Let me know if you want me to tackle those in this PR or handle them as separate, iterative PRs.

### Checklist

* [ ] Documentation has been updated *(N/A for this PR)*

### DIFF
--- a/src/docbuild/cli/cmd_build/__init__.py
+++ b/src/docbuild/cli/cmd_build/__init__.py
@@ -1,38 +1,4 @@
-"""CLI interface to build a document.
-
-A document (or "doctype") consists of ``[PRODUCT]/[DOCSET][@LIFECYCLES]/LANGS``
-with the following properties:
-
-\b
-
-* (Optional) ``PRODUCT`` is the product. To mark "ALL" products, omit the
-  product or use "*"
-* (Optional) ``DOCSET`` is the docset, usually the version or release of
-  a product.
-  To mark "ALL" docsets, omit the docset or use ``"*"``.
-* (Optional) ``LIFECYCLES`` marks a list of lifecycles separated by comma
-  or pipe symbol.
-  A lifecycle can be one of the values 'supported', 'unsupported', 'beta',
-  or 'hidden'.
-* ``LANGS`` marks a list of languages separated by comma. Every single
-  language contains a LANGUAGE-COUNTRY syntax, for example 'en-us', 'de-de' etc.
-
-Examples of the doctypes syntax:
-
-\b
-
-* ``"//en-us"``
-  Builds all supported deliverables for English
-* ``"sles/*/en-us"``
-  Builds only SLES deliverables which are supported and in English
-* ``"sles/*@unsupported/en-us,de-de"``
-  Builds all English and German SLES releases which are unsupported
-* ``"sles/@beta|supported/de-de"``
-  Build all docsets that are supported and beta for German SLES.
-* ``"sles/@beta,supported/de-de"``
-  Same as the previous one, but with comma as the separator between
-  the lifecycle states.
-"""  # noqa: D301
+"""CLI interface to build a document."""
 
 import asyncio
 from pathlib import Path
@@ -55,7 +21,7 @@ from ..context import DocBuildContext
 )
 @click.option(
     "-M", "--skip-metadata", is_flag=True, default=False, show_default=True,
-    help="Skip generating metadata after a successful build.",
+    help="Skip generating metadata (runs before build by default).",
 )
 @click.pass_context
 @requires_system_tools()
@@ -76,34 +42,35 @@ def build(
     assert context.envconfig is not None
     env = context.envconfig
 
-    # Paths for Build
+    # Paths for Metadata & Build
     main_portal_config = Path(env.paths.main_portal_config)
     repo_dir = Path(env.paths.repo_dir)
-    tmp_build_dir = Path(env.paths.tmp.tmp_build_base_dir) / str(env.paths.tmp.tmp_build_dir_dyn)
-
-    # Paths for Metadata
-    tmp_metadata_dir = Path(env.paths.tmp.tmp_metadata_dir)
     tmp_repo_dir = Path(env.paths.tmp_repo_dir)
+    tmp_metadata_dir = Path(env.paths.tmp.tmp_metadata_dir)
     meta_cache_dir = Path(env.paths.meta_cache_dir)
     json_cache_dir = Path(env.paths.json_cache_dir)
+    tmp_build_base_dir = Path(env.paths.tmp.tmp_build_base_dir)
+
     dapsmetatmpl = str(env.build.daps.meta)
+
+    # Dynamic templates for the build runner
+    # We use getattr safely until we formally add these to the EnvBuildDaps model
+    daps_tmpls = {
+        "html": getattr(env.build.daps, "html", "daps -d {{dcfile}} --builddir {{builddir}} html"),
+        "pdf": getattr(env.build.daps, "pdf", "daps -d {{dcfile}} --builddir {{builddir}} pdf"),
+        "single-html": getattr(env.build.daps, "single_html", "daps -d {{dcfile}} --builddir {{builddir}} single-html"),
+        "epub": getattr(env.build.daps, "epub", "daps -d {{dcfile}} --builddir {{builddir}} epub"),
+    }
 
     max_workers = context.appconfig.max_workers if context.appconfig else 1
 
     async def run_pipeline() -> int:
-        click.echo(f"[BUILD] Starting async build pipeline with {max_workers} workers...")
-        build_result = await build_process(
-            main_portal_config=main_portal_config,
-            repo_dir=repo_dir,
-            tmp_build_dir=tmp_build_dir,
-            max_workers=max_workers,
-            doctypes=doctypes,
-            skip_repo_update=skip_repo_update,
-        )
+        build_skip_repo = skip_repo_update
 
-        if build_result == 0 and not skip_metadata:
-            click.echo("[BUILD] Build successful. Chaining metadata generation...")
-            return await metadata_process(
+        # 1. Run Metadata (Fail-Fast Validation)
+        if not skip_metadata:
+            click.echo("[BUILD] Running metadata generation (fail-fast validation)...")
+            meta_result = await metadata_process(
                 main_portal_config=main_portal_config,
                 tmp_metadata_dir=tmp_metadata_dir,
                 repo_dir=repo_dir,
@@ -113,8 +80,28 @@ def build(
                 dapsmetatmpl=dapsmetatmpl,
                 max_workers=max_workers,
                 doctypes=list(doctypes),
-                skip_repo_update=True, # Repos were already updated by the build task
+                skip_repo_update=skip_repo_update,
             )
+
+            if meta_result != 0:
+                click.echo("[BUILD] Metadata generation failed. Aborting build.", err=True)
+                return meta_result
+
+            # Since repos were updated during metadata, skip updating them again during build
+            build_skip_repo = True
+
+        # 2. Run Build
+        click.echo(f"[BUILD] Starting async build pipeline with {max_workers} workers...")
+        build_result = await build_process(
+            main_portal_config=main_portal_config,
+            repo_dir=repo_dir,
+            tmp_build_base_dir=tmp_build_base_dir,
+            max_workers=max_workers,
+            doctypes=doctypes,
+            daps_tmpls=daps_tmpls,
+            skip_repo_update=build_skip_repo,
+        )
+
         return build_result
 
     result = asyncio.run(run_pipeline())
--- a/src/docbuild/cli/cmd_build/__init__.py
+++ b/src/docbuild/cli/cmd_build/__init__.py
@@ -1,16 +1,23 @@
 """CLI interface to build a document."""
 
 import asyncio
+import math
 from pathlib import Path
 
 import click
+from rich.console import Console
 
 from ...models.doctype import Doctype
 from ...tasks.build.runner import process as build_process
 from ...tasks.metadata.runner import process as metadata_process
+from ...utils.contextmgr import make_timer
 from ...utils.sysdeps import requires_system_tools
 from ..callback import validate_doctypes
 from ..context import DocBuildContext
+
+# Set up rich consoles for output
+stdout = Console()
+console_err = Console(stderr=True, style="red")
 
 
 @click.command(help="Subcommand to build a document.")
@@ -36,11 +43,14 @@ def build(
     context: DocBuildContext = ctx.obj
 
     if not context.envconfig:
-        click.echo("Environment configuration is missing.", err=True)
+        console_err.print("Environment configuration is missing.")
         ctx.exit(1)
 
     assert context.envconfig is not None
     env = context.envconfig
+
+    timer = make_timer("build")
+    result = 1  # Default exit code for interruption or error
 
     # Paths for Metadata & Build
     main_portal_config = Path(env.paths.main_portal_config)
@@ -52,9 +62,6 @@ def build(
     tmp_build_base_dir = Path(env.paths.tmp.tmp_build_base_dir)
 
     dapsmetatmpl = str(env.build.daps.meta)
-
-    # Dynamic templates for the build runner
-    # We use getattr safely until we formally add these to the EnvBuildDaps model
     daps_tmpls = {
         "html": getattr(env.build.daps, "html", "daps -d {{dcfile}} --builddir {{builddir}} html"),
         "pdf": getattr(env.build.daps, "pdf", "daps -d {{dcfile}} --builddir {{builddir}} pdf"),
@@ -63,13 +70,13 @@ def build(
     }
 
     max_workers = context.appconfig.max_workers if context.appconfig else 1
+    stdout.print(f"Config path: {env.paths.config_dir}")
 
     async def run_pipeline() -> int:
         build_skip_repo = skip_repo_update
 
-        # 1. Run Metadata (Fail-Fast Validation)
         if not skip_metadata:
-            click.echo("[BUILD] Running metadata generation (fail-fast validation)...")
+            stdout.print("[bold blue]Running metadata generation (fail-fast validation)...[/bold blue]")
             meta_result = await metadata_process(
                 main_portal_config=main_portal_config,
                 tmp_metadata_dir=tmp_metadata_dir,
@@ -84,15 +91,13 @@ def build(
             )
 
             if meta_result != 0:
-                click.echo("[BUILD] Metadata generation failed. Aborting build.", err=True)
+                console_err.print("Metadata generation failed. Aborting build.")
                 return meta_result
 
-            # Since repos were updated during metadata, skip updating them again during build
             build_skip_repo = True
 
-        # 2. Run Build
-        click.echo(f"[BUILD] Starting async build pipeline with {max_workers} workers...")
-        build_result = await build_process(
+        stdout.print(f"[bold blue]Starting async build pipeline with {max_workers} workers...[/bold blue]")
+        return await build_process(
             main_portal_config=main_portal_config,
             repo_dir=repo_dir,
             tmp_repo_dir=tmp_repo_dir,
@@ -103,7 +108,12 @@ def build(
             skip_repo_update=build_skip_repo,
         )
 
-        return build_result
+    t = None
+    try:
+        with timer() as t:
+            result = asyncio.run(run_pipeline())
+    finally:
+        if t and not math.isnan(t.elapsed):
+            stdout.print(f"Elapsed time {t.elapsed:0.2f}s")
 
-    result = asyncio.run(run_pipeline())
     ctx.exit(result)
--- a/src/docbuild/cli/cmd_build/__init__.py
+++ b/src/docbuild/cli/cmd_build/__init__.py
@@ -95,6 +95,7 @@ def build(
         build_result = await build_process(
             main_portal_config=main_portal_config,
             repo_dir=repo_dir,
+            tmp_repo_dir=tmp_repo_dir,
             tmp_build_base_dir=tmp_build_base_dir,
             max_workers=max_workers,
             doctypes=doctypes,
--- a/src/docbuild/cli/cmd_build/__init__.py
+++ b/src/docbuild/cli/cmd_build/__init__.py
@@ -40,39 +40,32 @@ from pathlib import Path
 import click
 
 from ...models.doctype import Doctype
-from ...tasks.build.runner import process
+from ...tasks.build.runner import process as build_process
+from ...tasks.metadata.runner import process as metadata_process
 from ...utils.sysdeps import requires_system_tools
 from ..callback import validate_doctypes
 from ..context import DocBuildContext
 
 
-@click.command(
-    help=__doc__.replace("\b\n\n", "\b\n").replace("``", ""),  # type: ignore
-)
-@click.argument(
-    "doctypes",
-    nargs=-1,
-    callback=validate_doctypes,
+@click.command(help="Subcommand to build a document.")
+@click.argument("doctypes", nargs=-1, callback=validate_doctypes)
+@click.option(
+    "-S", "--skip-repo-update", is_flag=True, default=False, show_default=True,
+    help="Skip updating git repositories before processing.",
 )
 @click.option(
-    "-S",
-    "--skip-repo-update",
-    is_flag=True,
-    default=False,
-    show_default=True,
-    help="Skip updating git repositories before processing.",
+    "-M", "--skip-metadata", is_flag=True, default=False, show_default=True,
+    help="Skip generating metadata after a successful build.",
 )
 @click.pass_context
 @requires_system_tools()
 def build(
-    ctx: click.Context, doctypes: tuple[Doctype, ...], skip_repo_update: bool
+    ctx: click.Context,
+    doctypes: tuple[Doctype, ...],
+    skip_repo_update: bool,
+    skip_metadata: bool,
 ) -> None:
-    """Subcommand build.
-
-    :param ctx: The Click context object.
-    :param doctypes: A tuple of doctype objects to build.
-    :param skip_repo_update: If True, do not fetch updates for the git repositories.
-    """
+    """Subcommand build."""
     ctx.ensure_object(DocBuildContext)
     context: DocBuildContext = ctx.obj
 
@@ -81,22 +74,48 @@ def build(
         ctx.exit(1)
 
     assert context.envconfig is not None
-
     env = context.envconfig
+
+    # Paths for Build
     main_portal_config = Path(env.paths.main_portal_config)
     repo_dir = Path(env.paths.repo_dir)
+    tmp_build_dir = Path(env.paths.tmp.tmp_build_base_dir) / str(env.paths.tmp.tmp_build_dir_dyn)
+
+    # Paths for Metadata
+    tmp_metadata_dir = Path(env.paths.tmp.tmp_metadata_dir)
+    tmp_repo_dir = Path(env.paths.tmp_repo_dir)
+    meta_cache_dir = Path(env.paths.meta_cache_dir)
+    json_cache_dir = Path(env.paths.json_cache_dir)
+    dapsmetatmpl = str(env.build.daps.meta)
 
     max_workers = context.appconfig.max_workers if context.appconfig else 1
 
-    click.echo(f"[BUILD] Starting async build pipeline with {max_workers} workers...")
-
-    result = asyncio.run(
-        process(
+    async def run_pipeline() -> int:
+        click.echo(f"[BUILD] Starting async build pipeline with {max_workers} workers...")
+        build_result = await build_process(
             main_portal_config=main_portal_config,
             repo_dir=repo_dir,
+            tmp_build_dir=tmp_build_dir,
             max_workers=max_workers,
             doctypes=doctypes,
             skip_repo_update=skip_repo_update,
         )
-    )
+
+        if build_result == 0 and not skip_metadata:
+            click.echo("[BUILD] Build successful. Chaining metadata generation...")
+            return await metadata_process(
+                main_portal_config=main_portal_config,
+                tmp_metadata_dir=tmp_metadata_dir,
+                repo_dir=repo_dir,
+                tmp_repo_dir=tmp_repo_dir,
+                meta_cache_dir=meta_cache_dir,
+                json_cache_dir=json_cache_dir,
+                dapsmetatmpl=dapsmetatmpl,
+                max_workers=max_workers,
+                doctypes=list(doctypes),
+                skip_repo_update=True, # Repos were already updated by the build task
+            )
+        return build_result
+
+    result = asyncio.run(run_pipeline())
     ctx.exit(result)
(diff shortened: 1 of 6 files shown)


tomschr:
This is great news! :tada: :+1: 

To answer your questions:

1. _Output Dirs: Should we adjust the DAPS command to output to a specific temporary directory (Step 6)?_

   I see two possible solutions to this:

   1. **No change in the daps call.** You build the deliverables as usual, for example `daps -d DC-SLE-HA-installation html`. That will create a `build` subdirectory in the respective source directory. That is together where the sources live (the work tree under ``).
     However, you need to _know exactly_ what's the output directory. Due to the async nature, other build jobs can write to the same `build` subdirectory.
     For example, in our case, the chunked HTML would be `build/SLE-HA-installation/html/article-installation/`.

   2. **Using `--builddir` option.** With this option, you can determine your own output directory. A good option would be pass the base temp directory from the ENV config using `paths.tmp_build_base_dir`. Maybe concat it with `paths.tmp.tmp_build_dir_dyn` to make it unique. Daps requires that the directory exists before you call it.
     In that case, daps will use this directory. If the directory path is unique, no other build jobs will write there.

  => Conclusion: From my gut feeling, I'd tend to option 2 as it's cleaner and safer.


2. _Chaining: Should the build command automatically trigger the metadata task upon a successful build (Steps 7-10)?_

   Yes, as a default. Perhaps we should add a CLI option `--skip-metadata` or something similar, if we want to separate the build process from the metadata process.
   Use case: you have to build a specific doctype several times because a writer waits for feedback from developers. In this case, you wouldn't want the metadata to slow down the process.
   On the other hand, on a regular "doc build day", you want to have the full chain.

3. _Sync: Where should the final sync/publish step live (Step 11)?_

   That should go to the value ´paths.target.target_base_dir` from the ENV config. 

   Maybe we need to implement issue #452 first.

   The sync could be a bit tricky. Docserv requires the following tree:
  
   ```
   SERVER_ROOT/
   └── LANGUAGE/
       └── PRODUCT/
           └── DOCSET/
               ├── html/
               ├── index.html
               ├── pdf/
               └── ARCHIVE.zip
   ```
   
   If you can arrange the temporary directory in a way that it resembles this structure, you only need to call `rsync SOURCE TARGET` and you are done.

sushant-suse:
> This is great news! 🎉 👍
> 
> To answer your questions:
> 
> 1. _Output Dirs: Should we adjust the DAPS command to output to a specific temporary directory (Step 6)?_
>    I see two possible solutions to this:
>    
>    1. **No change in the daps call.** You build the deliverables as usual, for example `daps -d DC-SLE-HA-installation html`. That will create a `build` subdirectory in the respective source directory. That is together where the sources live (the work tree under ``).
>       However, you need to _know exactly_ what's the output directory. Due to the async nature, other build jobs can write to the same `build` subdirectory.
>       For example, in our case, the chunked HTML would be `build/SLE-HA-installation/html/article-installation/`.
>    2. **Using `--builddir` option.** With this option, you can determine your own output directory. A good option would be pass the base temp directory from the ENV config using `paths.tmp_build_base_dir`. Maybe concat it with `paths.tmp.tmp_build_dir_dyn` to make it unique. Daps requires that the directory exists before you call it.
>       In that case, daps will use this directory. If the directory path is unique, no other build jobs will write there.
> 
> => Conclusion: From my gut feeling, I'd tend to option 2 as it's cleaner and safer.
> 
> 2. _Chaining: Should the build command automatically trigger the metadata task upon a successful build (Steps 7-10)?_
>    Yes, as a default. Perhaps we should add a CLI option `--skip-metadata` or something similar, if we want to separate the build process from the metadata process.
>    Use case: you have to build a specific doctype several times because a writer waits for feedback from developers. In this case, you wouldn't want the metadata to slow down the process.
>    On the other hand, on a regular "doc build day", you want to have the full chain.
> 3. _Sync: Where should the final sync/publish step live (Step 11)?_
>    That should go to the value ´paths.target.target_base_dir` from the ENV config.
>    Maybe we need to implement issue [Implement rsync capabilities #452](https://github.com/openSUSE/docbuild/issues/452) first.
>    The sync could be a bit tricky. Docserv requires the following tree:
>    ```
>    SERVER_ROOT/
>    └── LANGUAGE/
>        └── PRODUCT/
>            └── DOCSET/
>                ├── html/
>                ├── index.html
>                ├── pdf/
>                └── ARCHIVE.zip
>    ```
>    
>    
>        
>          
>        
>    
>          
>        
>    
>        
>      
>    If you can arrange the temporary directory in a way that it resembles this structure, you only need to call `rsync SOURCE TARGET` and you are done.

Thanks for the detailed breakdown, Toms. Option 2 for the build directories and adding a `--skip-metadata` flag makes perfect sense.

Here is my plan to wrap up this PR:

1. **Output Dirs:** I will update the `daps` execution to use `--builddir`, utilizing `paths.tmp_build_base_dir` + `paths.tmp.tmp_build_dir_dyn`, and ensure the directory is created before calling DAPS.
2. **Chaining:** I will add the `--skip-metadata` CLI option. By default, a successful build will now seamlessly invoke the metadata generation pipeline.

Regarding Step 3 (Syncing), I completely agree. It is much cleaner to pause the sync step here and tackle the new `rsync` issue (#452) first. Once that is in place, we can easily pipe the temporary tree to the `SERVER_ROOT`.

I'll get these updates pushed to this branch shortly.
Thanks.

src/docbuild/cli/cmd_build/__init__.py
```diff
@@ -50,18 +54,49 @@
...
     callback=validate_doctypes,
 )
+@click.option(
+    "-S",
+    "--skip-repo-update",
+    is_flag=True,
+    default=False,
+    show_default=True,
+    help="Skip updating git repositories before processing.",
+)
```
tomschr: Perhaps a `--skip-metadata` would be helpful too. See my last comment.


```suggestion
)
@click.option(
   # "-M",
   "--skip-metadata",
    is_flag=True,
    default=False,
    show_default=True,
    help="Skip creating metadata.",
)
```

src/docbuild/cli/cmd_build/__init__.py
```diff
@@ -50,18 +54,49 @@
...
 @click.pass_context
 @requires_system_tools()
-def build(ctx: click.Context, doctypes: tuple[Doctype]) -> None:
+def build(
+    ctx: click.Context, doctypes: tuple[Doctype, ...], skip_repo_update: bool
```
tomschr:
```suggestion
    ctx: click.Context, doctypes: tuple[Doctype, ...],
    skip_repo_update: bool,
    skip_metadata: bool,
```

src/docbuild/tasks/build/runner.py
```diff
@@ -0,0 +1,135 @@
...
+    doctype: Doctype,
+    repo_dir: Path,
+    max_workers: int,
+    *,
+    skip_repo_update: bool = False,
+) -> list[Deliverable]:
+    """Process a doctype and build its deliverables using aiostream."""
+    deliverables: list[Deliverable] = await asyncio.to_thread(
+        get_deliverable_from_doctype, root, doctype
+    )
```
tomschr: I think here you need to filter the list of deliverables. We only need deliverables of type "dc". In other words, we have to exclude the prebuilts.


```suggestion
    deliverables: list[Deliverable] = await asyncio.to_thread(
        get_deliverable_from_doctype, root, doctype
    )
    
    deliverables = [
        deli for deli in deliverables
        if deli.xml.is_dc
    ]
```

src/docbuild/tasks/build/runner.py
```diff
@@ -0,0 +1,144 @@
...
+    deliverable: Deliverable,
+    fmt: Literal["html", "pdf", "single-html", "epub"],
+    cwd: Path,
+    build_dir: Path,
+) -> tuple[bool, str]:
+    """Execute the DAPS build command for a specific format."""
+    dcfile = deliverable.xml.dcfile
+    if not dcfile:
+        log.warning("No DC file found for %s, skipping %s build.", deliverable.full_id, fmt)
+        return False, ""
```
tomschr: IMHO, the warning is unneccessary for two reasons:

* The validation already rejects invalid structures.
* If you apply my suggestion from a different comment, we would have filtered all deliverables to "DC" types only.

In other words, I'd suggest this change:


```suggestion
```

src/docbuild/tasks/build/runner.py
```diff
@@ -0,0 +1,144 @@
...
+
+    # Ensure unique output directory exists before calling daps
+    build_dir.mkdir(parents=True, exist_ok=True)
+
+    args = ["daps", "-d", dcfile, "--builddir", str(build_dir), fmt]
```
tomschr: I guess we should introduce a new entry in the ENV config. Probably under `build.daps.html` and `build.daps.pdf`? Similar what we did with `meta`.

That would make the logic slightly more difficult, but also more flexible. You would need to introduce some variables like `builddir`, `dcfile` etc. that you can replace to build the real command line.

src/docbuild/cli/cmd_build/__init__.py
```diff
@@ -34,34 +34,88 @@
...
+    main_portal_config = Path(env.paths.main_portal_config)
+    repo_dir = Path(env.paths.repo_dir)
+    tmp_build_dir = Path(env.paths.tmp.tmp_build_base_dir) / str(env.paths.tmp.tmp_build_dir_dyn)
+
+    # Paths for Metadata
+    tmp_metadata_dir = Path(env.paths.tmp.tmp_metadata_dir)
+    tmp_repo_dir = Path(env.paths.tmp_repo_dir)
+    meta_cache_dir = Path(env.paths.meta_cache_dir)
+    json_cache_dir = Path(env.paths.json_cache_dir)
+    dapsmetatmpl = str(env.build.daps.meta)
```
tomschr, line 101: Aren't they already `Path`-like objects? If not, shouldn't they be after validation?

We have this `EnsureWritableDirectory` class, but that's not a `Path` unfortunately. I think, this kind of conversion should be hidden behind a Pydantic model or some other class.

What's the best option here? Modify `EnsureWritableDirectory` to be derived from `Path`? Adjust some Pydantic models when creating the ENV context?

src/docbuild/cli/cmd_build/__init__.py
```diff
@@ -34,34 +34,88 @@
...
+                repo_dir=repo_dir,
+                tmp_repo_dir=tmp_repo_dir,
+                meta_cache_dir=meta_cache_dir,
+                json_cache_dir=json_cache_dir,
+                dapsmetatmpl=dapsmetatmpl,
+                max_workers=max_workers,
+                doctypes=list(doctypes),
+                skip_repo_update=True, # Repos were already updated by the build task
+            )
+        return build_result
```
tomschr: Actually, I would actually switch order of these: call the metadata first and then the real build.

The reason for this is: if something is wrong, a `daps metadata` would fail fast. That means, you don't have to start the build process.

src/docbuild/tasks/build/runner.py
```diff
@@ -0,0 +1,144 @@
...
+    cwd = repo_dir / deliverable.subdir if deliverable.subdir else repo_dir
+
+    # Create a completely unique output path to prevent concurrent build collisions
+    safe_id = deliverable.full_id.replace("/", "_").replace(":", "_")
+    deliverable_build_dir = tmp_build_dir / safe_id
+
+    success = True
+    for fmt, is_enabled in deliverable.format.items():
+        if is_enabled:
+            fmt_success, _ = await build_format(deliverable, fmt, cwd, deliverable_build_dir)
```
tomschr: To really create a safe directory (safe = avoiding race conditions), you need to could use the `PersistentOnErrorTemporaryDirectory` class which uses the `tempfile.TemporaryDirectory` as a base classe. It creates a temporary path like this:

```
unique_temp_path= dir + prefix + random_chars + suffix
```

The random characters are created by `tempfile`. Additionally, it deletes the temporary directory only if no exception occurs within the context block.

```suggestion
    safe_id = deliverable.make_safe_name(deliverable.full_id)

    success = True
    for fmt, is_enabled in deliverable.format.items():
        if is_enabled:
            with PersistentOnErrorTemporaryDirectory(
                dir=tmp_build_base_dir,
                prefix=f"{safe_id}_",
                suffix=f"_{fmt}",
                # delete = True|False # Wheather the directory is automatically deleted
            ) as deliverable_build_dir:
                fmt_success, _ = await build_format(deliverable, fmt, cwd, deliverable_build_dir)
```

The

src/docbuild/tasks/build/runner.py
```diff
@@ -0,0 +1,144 @@
...
+    tmp_build_dir: Path,
+    max_workers: int,
+    *,
+    skip_repo_update: bool = False,
+) -> list[Deliverable]:
+    """Process a doctype and build its deliverables using aiostream."""
+    deliverables: list[Deliverable] = await asyncio.to_thread(
+        get_deliverable_from_doctype, root, doctype
+    )
+    deliverables.sort()
```
tomschr, line 125: I think here you need to filter the list of deliverables. We only need deliverables of type "dc". Prebuilt deliverables can't be built, so we need to exclude them.

```suggestion
    deliverables: list[Deliverable] = await asyncio.to_thread(
        get_deliverable_from_doctype, root, doctype
    )
    deliverables = [
        deli for deli in deliverables
        if deli.xml.is_dc
    ]
    deliverables.sort()
```

tomschr (changes requested):
That's a great start! 🎉 👍 Thank you very much!

I have some suggestions. Let me know if you have any questions.

When I try to build it, I get some errors:

```
$ docbuild -vv build --skip-metadata 'suse-ai/1.0/en-us'
warning: `VIRTUAL_ENV=/home/toms/repos/GH/opensuse/toms/doc-doctype/.venv` does not match the project environment path `.venv` and will be ignored; use `--active` to target the active environment instead
[2026-08-19 12:20:44] [DEBUG] [None] asyncio: Using selector: EpollSelector
[BUILD] Starting async build pipeline with 8 workers...
[2026-08-19 12:20:44] [INFO] [Task-2] docbuild.tasks.metadata.repos: Updating Git repositories...
[2026-08-19 12:20:44] [INFO] [Task-3] docbuild.git: Repository already exists, fetching updates for 'suse/doc-suse-ai'
[2026-08-19 12:20:44] [INFO] [Task-3] docbuild.git: Trying to fetch updates for 'https___github_com_suse_doc_suse_ai_git'
[2026-08-19 12:20:44] [DEBUG] [Task-3] docbuild.git: Executing Git command: git -c color.ui=never fetch origin +refs/heads/*:refs/heads/* -v --prune in /home/toms/.local/state/docbuild/repos/permanent/https___github_com_suse_doc_suse_ai_git
[2026-08-19 12:20:46] [INFO] [Task-3] docbuild.git: Successfully fetched updates for 'https___github_com_suse_doc_suse_ai_git'
[2026-08-19 12:20:46] [INFO] [Task-2] docbuild.tasks.build.runner: Building html for suse-ai/1.0/main/en-us:DC-AI-deployment...
[2026-08-19 12:20:46] [ERROR] [Task-2] docbuild.tasks.build.runner: Failed to build html for suse-ai/1.0/main/en-us:DC-AI-deployment:
ERROR: Wrong path for the DC-file
       (provided via config file: DOCCONF=DC-AI-deployment)
[2026-08-19 12:20:46] [INFO] [Task-2] docbuild.tasks.build.runner: Building pdf for suse-ai/1.0/main/en-us:DC-AI-deployment...
[2026-08-19 12:20:46] [ERROR] [Task-2] docbuild.tasks.build.runner: Failed to build pdf for suse-ai/1.0/main/en-us:DC-AI-deployment:
ERROR: Wrong path for the DC-file
       (provided via config file: DOCCONF=DC-AI-deployment)
[2026-08-19 12:20:46] [ERROR] [Task-1] docbuild.tasks.build.runner: Build completed with 1 failures.
```

sushant-suse:
> That's a great start! 🎉 👍 Thank you very much!
> 
> I have some suggestions. Let me know if you have any questions.
> 
> When I try to build it, I get some errors:
> 
> ```
> $ docbuild -vv build --skip-metadata 'suse-ai/1.0/en-us'
> warning: `VIRTUAL_ENV=/home/toms/repos/GH/opensuse/toms/doc-doctype/.venv` does not match the project environment path `.venv` and will be ignored; use `--active` to target the active environment instead
> [2026-08-19 12:20:44] [DEBUG] [None] asyncio: Using selector: EpollSelector
> [BUILD] Starting async build pipeline with 8 workers...
> [2026-08-19 12:20:44] [INFO] [Task-2] docbuild.tasks.metadata.repos: Updating Git repositories...
> [2026-08-19 12:20:44] [INFO] [Task-3] docbuild.git: Repository already exists, fetching updates for 'suse/doc-suse-ai'
> [2026-08-19 12:20:44] [INFO] [Task-3] docbuild.git: Trying to fetch updates for 'https___github_com_suse_doc_suse_ai_git'
> [2026-08-19 12:20:44] [DEBUG] [Task-3] docbuild.git: Executing Git command: git -c color.ui=never fetch origin +refs/heads/*:refs/heads/* -v --prune in /home/toms/.local/state/docbuild/repos/permanent/https___github_com_suse_doc_suse_ai_git
> [2026-08-19 12:20:46] [INFO] [Task-3] docbuild.git: Successfully fetched updates for 'https___github_com_suse_doc_suse_ai_git'
> [2026-08-19 12:20:46] [INFO] [Task-2] docbuild.tasks.build.runner: Building html for suse-ai/1.0/main/en-us:DC-AI-deployment...
> [2026-08-19 12:20:46] [ERROR] [Task-2] docbuild.tasks.build.runner: Failed to build html for suse-ai/1.0/main/en-us:DC-AI-deployment:
> ERROR: Wrong path for the DC-file
>        (provided via config file: DOCCONF=DC-AI-deployment)
> [2026-08-19 12:20:46] [INFO] [Task-2] docbuild.tasks.build.runner: Building pdf for suse-ai/1.0/main/en-us:DC-AI-deployment...
> [2026-08-19 12:20:46] [ERROR] [Task-2] docbuild.tasks.build.runner: Failed to build pdf for suse-ai/1.0/main/en-us:DC-AI-deployment:
> ERROR: Wrong path for the DC-file
>        (provided via config file: DOCCONF=DC-AI-deployment)
> [2026-08-19 12:20:46] [ERROR] [Task-1] docbuild.tasks.build.runner: Build completed with 1 failures.
> ```

Thanks so much for the review and for testing it out, Toms 🎉

Ah, I see the `Wrong path for the DC-file` error—it looks like `daps` is either firing in the wrong `cwd` or we need to ensure the correct worktree is checked out before building. I'll go through your line-by-line comments right now and get these fixed up.

src/docbuild/tasks/build/runner.py
sushant-suse: Removed the unnecessary dcfile check.

src/docbuild/tasks/build/runner.py
sushant-suse: Replaced the static temp directory logic with `PersistentOnErrorTemporaryDirectory` and `deliverable.make_safe_name()` to guarantee isolated, collision-free build environments.

src/docbuild/tasks/build/runner.py
sushant-suse: Filtered the `deliverables` list using `if deli.xml.is_dc`

src/docbuild/cli/cmd_build/__init__.py
sushant-suse: Great observation. `EnsureWritableDirectory` currently behaves as a string-like type at runtime, which is why the `Path()` wrapper is needed here. The best long-term fix would be to refactor `EnsureWritableDirectory` to inherit from `pathlib.Path` (or use a Pydantic `BeforeValidator` to coerce it). Since that touches the global config models, I think we should tackle that in a separate PR to keep this one focused. I'll leave the wrappers for now.

src/docbuild/cli/cmd_build/__init__.py
sushant-suse: That is a brilliant idea. Using the metadata step as a fail-fast validation makes total sense. I have swapped the execution order so `metadata` runs before `build`, and configured it so the build step skips the git update if the metadata step already fetched the repos.

src/docbuild/cli/cmd_build/__init__.py
tomschr: > Since that touches the global config models, I think we should tackle that in a separate PR to keep this one focused. I'll leave the wrappers for now.

Yes, agree with you. I was just wondering. Please open an issue for that part. Good thing is, we can adjust it everywhere. 😉

src/docbuild/cli/cmd_build/__init__.py
```diff
@@ -1,67 +1,109 @@
...
-* ``"sles/*/en-us"``
-  Builds only SLES deliverables which are supported and in English
-* ``"sles/*@unsupported/en-us,de-de"``
-  Builds all English and German SLES releases which are unsupported
-* ``"sles/@beta|supported/de-de"``
-  Build all docsets that are supported and beta for German SLES.
-* ``"sles/@beta,supported/de-de"``
-  Same as the previous one, but with comma as the separator between
-  the lifecycle states.
-"""  # noqa: D301
```
tomschr, line 35: If possible, I would keep that as it serves as a) module documentation, and b) help for the CLI subcommand.

src/docbuild/cli/cmd_build/__init__.py
```diff
@@ -1,67 +1,109 @@
...
+from ...tasks.build.runner import process as build_process
+from ...tasks.metadata.runner import process as metadata_process
 from ...utils.sysdeps import requires_system_tools
 from ..callback import validate_doctypes
 from ..context import DocBuildContext
 
 
-@click.command(
-    help=__doc__.replace("\b\n\n", "\b\n").replace("``", ""),  # type: ignore
+@click.command(help="Subcommand to build a document.")
```
tomschr: I would keep the original one (with the `help=__doc__` part).

src/docbuild/cli/cmd_build/__init__.py
tomschr, line 1: If you change the defaults for our env config due to formats, adjust `docserv.cli.defaults.DEFAULT_ENV_CONFIG` accordingly.

src/docbuild/models/config/env.py
```diff
@@ -61,6 +61,34 @@ class EnvBuildDaps(BaseModel):
     )
     "The command used to extract DAPS metadata."
 
+    html: str = Field(
+        default="daps -d {{dcfile}} --builddir {{builddir}} html",
```
tomschr: I would make a small replacement with daps:


```suggestion
        default="{build.daps.command} -d {{dcfile}} --builddir {{builddir}} html",
```

The reason for this: if you want to switch from the system `daps` command to one from GitHub, you can change just the key in `build.daps.command` and all the other dependent commands adjust accordingly.

src/docbuild/models/config/env.py
```diff
@@ -61,6 +61,34 @@ class EnvBuildDaps(BaseModel):
...
+
+    single_html: str = Field(
+        default="daps -d {{dcfile}} --builddir {{builddir}} single-html",
+        title="DAPS Single HTML Command Template",
+        description="The template string to build Single HTML.",
+    )
+    "The command template used to build Single HTML."
+
+    epub: str = Field(
+        default="daps -d {{dcfile}} --builddir {{builddir}} epub",
```
tomschr: Same here:


```suggestion
        default="{build.daps.command} -d {{dcfile}} --builddir {{builddir}} pdf",
        title="DAPS PDF Command Template",
        description="The template string to build PDF.",
    )
    "The command template used to build PDF."

    single_html: str = Field(
        default="{build.daps.command} -d {{dcfile}} --builddir {{builddir}} single-html",
        title="DAPS Single HTML Command Template",
        description="The template string to build Single HTML.",
    )
    "The command template used to build Single HTML."

    epub: str = Field(
        default="{build.daps.command} -d {{dcfile}} --builddir {{builddir}} epub",
```

src/docbuild/tasks/build/runner.py
```diff
@@ -0,0 +1,185 @@
...
+                    prefix=f"build_{safe_id}_",
+                    suffix=f"_{fmt}",
+                ))
+
+                fmt_success, _ = await build_format(
+                    deliverable, fmt, cwd, deliverable_build_dir, tmpl
+                )
+                if not fmt_success:
+                    success = False
+
```
tomschr, line 104: Something important is missing here: you need to sync the results into the target directory. 😉 


```suggestion
                if not fmt_success:
                    success = False
                else:
                    # Call rsync to target directory
                    log.debug("Syncing result to ...")

```

tomschr (changes requested):
I found some other issue that I haven't thought about yet.
Thanks!

src/docbuild/cli/cmd_build/__init__.py
sushant-suse: I agree. Added them back.

src/docbuild/tasks/build/runner.py
sushant-suse: Oh yes, I missed that part. Thanks Toms.

src/docbuild/cli/cmd_build/__init__.py
sushant-suse: Noted sir

src/docbuild/cli/defaults.py
```diff
@@ -89,6 +89,10 @@
         "daps": {
             "command": "daps",
             "meta": "daps --builddir={{builddir}} -d {{dcfile}} metadata --output {{output}}",
+            "html": "{build.daps.command} -d {{dcfile}} --builddir {{builddir}} html",
+            "pdf": "{build.daps.command} -d {{dcfile}} --builddir {{builddir}} pdf",
+            "single_html": "{build.daps.command} -d {{dcfile}} --builddir {{builddir}} single-html",
+            "epub": "{build.daps.command} -d {{dcfile}} --builddir {{builddir}} epub",
```
tomschr: Just noticed: the `--builddir` is a _global_ option. As such, move it _after_ the daps command:


```suggestion
            "html": "{build.daps.command} --builddir {{builddir}} -d {{dcfile}}  html",
            "pdf": "{build.daps.command} --builddir {{builddir}} -d {{dcfile}} pdf",
            "single_html": "{build.daps.command} --builddir {{builddir}} -d {{dcfile}} single-html",
            "epub": "{build.daps.command} --builddir {{builddir}} -d {{dcfile}} epub",
```

src/docbuild/models/config/env.py
```diff
@@ -61,6 +61,34 @@ class EnvBuildDaps(BaseModel):
     )
     "The command used to extract DAPS metadata."
 
+    html: str = Field(
+        default="{build.daps.command} -d {{dcfile}} --builddir {{builddir}} html",
```
tomschr: Same here:


```suggestion
        default="{build.daps.command} --builddir {{builddir}} -d {{dcfile}} html",
```

src/docbuild/models/config/env.py
```diff
@@ -61,6 +61,34 @@ class EnvBuildDaps(BaseModel):
...
+        title="DAPS HTML Command Template",
+        description="The template string to build HTML.",
+    )
+    "The command template used to build HTML."
+
+    pdf: str = Field(
+        default="{build.daps.command} -d {{dcfile}} --builddir {{builddir}} pdf",
```
tomschr:
```suggestion
        default="{build.daps.command} --builddir {{builddir}} -d {{dcfile}} pdf",
```

src/docbuild/models/config/env.py
```diff
@@ -61,6 +61,34 @@ class EnvBuildDaps(BaseModel):
...
+        title="DAPS PDF Command Template",
+        description="The template string to build PDF.",
+    )
+    "The command template used to build PDF."
+
+    single_html: str = Field(
+        default="{build.daps.command} -d {{dcfile}} --builddir {{builddir}} single-html",
```
tomschr:
```suggestion
        default="{build.daps.command} --builddir {{builddir}} -d {{dcfile}} single-html",
```

src/docbuild/models/config/env.py
```diff
@@ -61,6 +61,34 @@ class EnvBuildDaps(BaseModel):
...
+        title="DAPS Single HTML Command Template",
+        description="The template string to build Single HTML.",
+    )
+    "The command template used to build Single HTML."
+
+    epub: str = Field(
+        default="{build.daps.command} -d {{dcfile}} --builddir {{builddir}} epub",
```
tomschr:
```suggestion
        default="{build.daps.command} --builddir {{builddir}} -d {{dcfile}}  epub",
```

src/docbuild/tasks/build/runner.py
```diff
@@ -0,0 +1,188 @@
...
+) -> tuple[bool, str]:
+    """Execute the DAPS build command dynamically from a config template."""
+    dcfile = deliverable.xml.dcfile
+    assert dcfile is not None, "Deliverable must have a DC file."
+
+    cmd_str = (
+        daps_tmpl.replace("{{dcfile}}", dcfile)
+        .replace("{{builddir}}", str(build_dir))
+        .replace("{{format}}", fmt)
+    )
```
tomschr: This is wrong and leads to the following error when calling `docbuild build`:

```
[2026-08-20 07:00:39] [ERROR] [Task-17] docbuild.tasks.build.runner: Failed to build html for suse-ai/1.0/main/en-us:DC-AI-deployment:
ERROR: BUILD_DIR "{builddir}" is not a valid directory
```

Using `.replace` doesn't lead to the expected replacement. In this case use the `.format` method:

```suggestion
    cmd_str = daps_tmpl.format(
        dcfile=dcfile,
        builddir=str(build_dir),
        format=fmt,
    )
```

src/docbuild/tasks/build/runner.py
tomschr: Keep in mind, that part needs to be replaced with the rsync from issue #452. Just saying. 😉

For your information: the old Docserv code collected all formats in a predefined temporary structure and synced it then to the target directory.

We don't have to follow this old strategy. Whatever works for us can be used.

tomschr (changes requested):
Great work, Sushant! 👍

I've tried `docbuild build` again and got a problem with some placeholders in `build.daps.html`. When I fix that, it runs perfectly (without the sync, but that's expected).

src/docbuild/tasks/build/runner.py
sushant-suse: Noted Toms

tomschr (approved):
Looks great! 👍 Thank you so much for all your efforts.

Ready to squash-merge it to main! 🚀

PR #460, opened 2026-08-20
feat(utils): implement async rsync wrapper with RsyncOptions

sushant-suse:
Addresses #452

**What was done:**

* **Created RsyncOptions Dataclass:** Implemented a flexible configuration class in `docbuild.utils.sync` to handle common rsync flags (`archive`, `delete`, `dry_run`, etc.) alongside an `extra_args` list, preventing function signature bloat while maintaining full CLI flexibility.
* **Implemented Async `rsync` Wrapper:** Created an asynchronous `rsync()` function that wraps `docbuild.utils.shell.run_command` to execute synchronization tasks safely and concurrently.
* **Smart Trailing Slash Handling:** Added path inspection to automatically infer if the user intends to sync a directory or just its contents based on the presence of a trailing slash, with an explicit `content_only` override parameter.
* **Registered System Dependency:** Added `rsync` to the `SYSTEM_DEPENDENCIES` dictionary in `constants.py` so it is automatically verified by the `docbuild doctor` command.
* **Added Comprehensive Tests:** Wrote isolated unit tests verifying argument generation, trailing slash behavior, and `Path` object compatibility.
* **Resolved Type Warnings:** Fixed a pedantic Pylance strict-typing error regarding `Enum.name` types in `constants.py`.

@tomschr, as we discussed, I kept this PR strictly focused on the core `rsync` utility implementation to keep a clean separation of concerns. Once this is merged, I will open the next PR to actually integrate this into the `build` subcommand logic.

### Checklist

* [ ] Documentation has been updated *(N/A for this PR)*

### DIFF
--- a/changelog.d/452.feature.rst
+++ b/changelog.d/452.feature.rst
@@ -1,1 +1,1 @@
-Added an async ``rsync`` utility function and ``RsyncOptions`` dataclass for flexible directory synchronization, and registered ``rsync`` as a validated system dependency.
+Added an async :command:`rsync` utility function and ``RsyncOptions`` dataclass for flexible directory synchronization, and registered :command:`rsync` as a validated system dependency.
--- a/changelog.d/452.feature.rst
+++ b/changelog.d/452.feature.rst
@@ -1,0 +1,1 @@
+Added an async ``rsync`` utility function and ``RsyncOptions`` dataclass for flexible directory synchronization, and registered ``rsync`` as a validated system dependency.
--- a/src/docbuild/constants.py
+++ b/src/docbuild/constants.py
@@ -39,7 +39,7 @@ SERVER_ROLES_ALIASES: tuple[str, ...] = tuple(ServerRole.__members__.keys())
 DEFAULT_LIFECYCLE: str = "supported"
 """The default lifecycle state for a docset."""
 
-ALLOWED_LIFECYCLES: tuple[str, ...] = tuple(lc.name for lc in LifecycleFlag)
+ALLOWED_LIFECYCLES: tuple[str, ...] = tuple(str(lc.name) for lc in LifecycleFlag if lc.name)
 # ('supported', 'beta', 'hidden', 'unsupported')
 """The available lifecycle states for a docset (without 'unknown')."""
 
@@ -170,4 +170,5 @@ SYSTEM_DEPENDENCIES = {
     "daps": ">=4",
     "xmllint": None,
     "xsltproc": None,
+    "rsync": None,
 }
--- a/src/docbuild/utils/sync.py
+++ b/src/docbuild/utils/sync.py
@@ -1,0 +1,89 @@
+"""Synchronization utilities."""
+
+from __future__ import annotations
+
+from dataclasses import dataclass, field
+import os
+from pathlib import Path
+import subprocess
+
+from .shell import run_command
+
+
+@dataclass(frozen=True, slots=True)
+class RsyncOptions:
+    """Configuration options for rsync execution."""
+
+    archive: bool = True
+    compress: bool = False
+    delete: bool = False
+    dry_run: bool = False
+    verbose: bool = False
+    partial: bool = False
+
+    # Acts as an escape hatch for the 100+ rsync options not explicitly modeled here.
+    # Users can pass arbitrary flags (e.g., ["--exclude=*.tmp", "--bwlimit=1000"]).
+    extra_args: list[str] = field(default_factory=list)
+
+    def to_args(self) -> list[str]:
+        """Convert the configured options into a list of command-line arguments.
+
+        :return: A list of string arguments formatted for the rsync command.
+        """
+        args: list[str] = []
+
+        if self.archive:
+            args.append("-a")
+        if self.compress:
+            args.append("-z")
+        if self.delete:
+            args.append("--delete")
+        if self.dry_run:
+            args.append("--dry-run")
+        if self.verbose:
+            args.append("-v")
+        if self.partial:
+            args.append("--partial")
+
+        args.extend(self.extra_args)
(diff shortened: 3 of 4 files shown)


changelog.d/452.feature.rst
```diff
@@ -0,0 +1 @@
+Added an async ``rsync`` utility function and ``RsyncOptions`` dataclass for flexible directory synchronization, and registered ``rsync`` as a validated system dependency.
```
tomschr:
```suggestion
Added an async :command:`rsync` utility function and ``RsyncOptions`` dataclass for flexible directory synchronization, and registered :command:`rsync` as a validated system dependency.
```

src/docbuild/utils/sync.py
```diff
@@ -0,0 +1,89 @@
...
+        if self.dry_run:
+            args.append("--dry-run")
+        if self.verbose:
+            args.append("-v")
+        if self.partial:
+            args.append("--partial")
+
+        args.extend(self.extra_args)
+
+        return args
```
tomschr: I've refined my original one to avoid the if-clauses and to define the CLI options directly in the type definitions. I think that is a better and more pythonic implementation than my old, naive idea. 😉

You can find it in the original issue.

src/docbuild/constants.py
```diff
@@ -39,7 +39,7 @@
 DEFAULT_LIFECYCLE: str = "supported"
 """The default lifecycle state for a docset."""
 
-ALLOWED_LIFECYCLES: tuple[str, ...] = tuple(lc.name for lc in LifecycleFlag)
+ALLOWED_LIFECYCLES: tuple[str, ...] = tuple(str(lc.name) for lc in LifecycleFlag if lc.name)
```
tomschr, line 42: The if clause is good, but I don't think we need `str` here, because:

```python
>>> from docbuild.models.lifecycle import LifecycleFlag
>>> [lc.name  for lc in LifecycleFlag if lc.name]
['supported', 'beta', 'hidden', 'unsupported']
```

tests/utils/test_sync.py
```diff
@@ -0,0 +1,70 @@
...
+    options = RsyncOptions(
+        archive=False,
+        compress=True,
+        delete=True,
+        dry_run=True,
+        verbose=True,
+        partial=True,
+        extra_args=["--exclude=*.tmp"]
+    )
+    assert options.to_args() == ["-z", "--delete", "--dry-run", "-v", "--partial", "--exclude=*.tmp"]
```
tomschr: This looks like you could combine the two test functions into one using Pytest's `parametrize` decorator:


```suggestion
@pytest.mark.parametrize(
    ("options", "expected"),
    [
        (
            RsyncOptions(),
            ["-a"],
        ),
        (
            RsyncOptions(
                archive=False,
                compress=True,
                delete=True,
                dry_run=True,
                verbose=True,
                partial=True,
                extra_args=["--exclude=*.tmp"],
            ),
            ["-z", "--delete", "--dry-run", "-v", "--partial", "--exclude=*.tmp"],
        ),
    ],
    ids=["defaults", "custom_flags"],
)
def test_rsync_options_to_args(options: RsyncOptions, expected: list[str]) -> None:
    assert options.to_args() == expected
```

tests/utils/test_sync.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+@pytest.mark.asyncio
+@patch("docbuild.utils.sync.run_command", new_callable=AsyncMock)
```
tomschr: Better use `patch.object` here as this is for VSCode easier to refactor.

```python
import docbuild.utils.sync as sync_mod

#...
@patch.object(sync_mod, "run_command", new_callable=AsyncMock)
```

tests/utils/test_sync.py
```diff
@@ -0,0 +1,70 @@
...
+
+    mock_run_command.reset_mock()
+
+    # Even with trailing slash in input, if False, it should NOT append
+    await rsync("source_dir/", "target_dir", content_only=False)
+    mock_run_command.assert_called_once_with(["rsync", "-a", "source_dir", "target_dir"])
+
+
+@pytest.mark.asyncio
+@patch("docbuild.utils.sync.run_command", new_callable=AsyncMock)
```
tomschr: Same with the other `@patch` decorators.

tomschr (changes requested):
Great work! 👍

I have some small comments for you.

src/docbuild/constants.py
sushant-suse: I added it because Pylance was showing error. Should I revert it back?

tests/utils/test_sync.py
sushant-suse: Great suggestion. Parametrizing this makes the test file much cleaner and easier to extend if we ever add more options. Done.

tests/utils/test_sync.py
sushant-suse: Ah, that makes perfect sense for IDE refactoring. I've updated all the tests in this file.

src/docbuild/utils/sync.py
sushant-suse: This is a brilliant use of `dataclasses.field` metadata and Python's match statements! You are totally right; it makes the class incredibly extensible and much more Pythonic. I have updated the implementation to use this pattern (and fixed a tiny duplicate `extra_args` typo in the snippet 😉). Pushing the update now.

src/docbuild/constants.py
tomschr: Oh really? Strange. I though the `.name` property is already str. Ok, then keep it.

tomschr (approved):
Looks great! Thank you very much, Sushant! 👍

Please merge it!

